Issue #2, opened 2018-03-05
Packaging script may need to include COPYING with docker images, docker, Drupal copyright info


mradcliffe:
## Feature request:

We may need to include a COPYING file with licenses and copyright for any software we distribute.

## What you expected to happen:

Packaging script should include COPYING file (or keep it in this repository) with the license for this software, ddev, docker, docker images and Drupal.

## Possible example text:

> This Software is distributed in a compressed archive with the following third-party software and Docker images under other copyright notices.
>
>   ddev is included in the distributable release as a software artifact, and is distributed under the Apache 2.0 license:
>
>      ddev - (c) Copyright 2016 DRUD Technology, LLC
> 
>   Docker images are included in the distributable release as software artifacts and distributed according to the provisions of the MIT license:
> 
>     drud/nginx-php-fpm-local - (c) Copyright 2017 DRUD Technology, LLC
>     drud/docker-phpmyadmin - (c) Copyright 2017 DRUD Technology, LLC
>     hoadx/cloud9-alpine - (c) Copyright 2017 hoadx
>     linuxserver/thelounge - (c) Copyright 2016 LinuxServer.io
>
>   Docker images are included in the distributable release as software artifacts and distributed according to the provisions of the Apache 2.0 license:
>
>     drud/ddev-router - (c) Copyright 2016 DRUD Technology, LLC
>     drud/mariadb-local - (c) Copyright 2016 DRUD Technology, LLC
>     mailhog/mailhog - (c) Copyright 2014 - 2017 Ian Kent (http://iankent.uk)
>
>  Drupal is included in the distributable release as a software artifact, and is distributed under the GPL license:
> 
>     Drupal - (c) Copyright 2001 - 2013 by the original authors
>       See drupal/core/COPYRIGHT.txt for more details.
> 
> This Software may be distributed alongside the following third-party software under other copyright notices.
> 
>   Docker Community Edition, or otherwise known as "Docker open source", is distributed under the Apache 2.0 license under the following copyright:
> 
>     Docker open source - (c) 2013 - 2017 Docker, Inc. 
>       See https://www.docker.com/components-licenses for more details.
>

rfay:
This is absolutely necessary, of course. PR is sure welcome.

mradcliffe:
Issue resolved.

PR #8, opened 2018-03-10
Adds licenses for docker images, docker ce and ddev and copyright info

mradcliffe:
Re: issue #2.

I am not a lawyer.


mradcliffe:
Also ping @rickmanelius

rickmanelius:
Reviewing now. I did a significant amount of work with our legal team, so I'll serve as a proxy and raise a question if necessary.

rickmanelius (changes requested):
Hi @mradcliffe. This is great! A few suggestions (I can fork and modify if you'd prefer).

* There's a small formatting difference in spacing for the different Apache2 licenses. Since they are all the same content-wise, it might make sense to fix that spacing or just copy. 
* I think we should prepend "LICENSE-" on all the files so that it's easier/more obvious as to what they are and people can feel free to visually skip by them or dive into each. Otherwise, as an end-user I would be confused as to what each file was until I viewed them each individually.

mradcliffe:
Good point on prepending LICENSE- to make it less confusing to a user as to what these text files are.

Talked to @rickmanelius about the formatting differences. I Think that's OK since those files are coming from their respective repositories.

rickmanelius (approved):
LGTM +1

rfay (approved):
This seems great to me. I'll pull it after the CLA is signed.

mradcliffe:
Closing for #11

rfay:
Next time it's probably easier to just amend the commits with the right email and force-push.

PR #12, opened 2018-03-11
Drupalcon based changes ( Mac OS, untested Linux, untested Windows)

BrianGilbert:
If Docker isn't already present lets install, and open it considering we're already packaging it.


rickmanelius:
+1 for this because while some people had it, others didn't and the scripts will fail without this in place.

BrianGilbert:
Added extraction change for xz based images package. Relies on v0.15.1 release of ddev.

install_ddev.sh
```diff
@@ -14,26 +14,39 @@ BINOWNER=$(ls -ld /usr/local/bin | awk '{print $3}')
 USER=$(whoami)
 SHACMD=""
 FILEBASE=""
+CURRENT_DIR=$PWD
 
 if [[ "$OS" == "Darwin" ]]; then
     FILEBASE="ddev_macos"
+
+    if ! docker --version >/dev/null 2>&1; then
```
rfay: Let's give people a chance here before just installing on their machine. Can we give them a Y/N option before proceeding? docker is great, but they need to agree to installing it. The choice could be here, or perhaps we could do it earlier.

Are you going to push this on Windows as well?

install_ddev.sh
```diff
@@ -14,26 +14,39 @@ BINOWNER=$(ls -ld /usr/local/bin | awk '{print $3}')
...
-    printf "${YELLOW}Docker is required for ddev. Download and install docker at https://www.docker.com/community-edition#/download before attempting to use ddev.${RESET}\n"
-fi
-
-if ! docker-compose --version >/dev/null 2>&1; then
-    printf "${YELLOW}Docker Compose is required for ddev. Download and install docker-compose at https://www.docker.com/community-edition#/download before attempting to use ddev.${RESET}\n"
-fi
-
 echo "Installing docker images for ddev to use..."
-gzip -dc $(ls ddev_tarballs/ddev_docker_images*.tar.gz) | docker load
+tar -xOpvf $(ls ddev_tarballs/ddev_docker_images*.tar.xz) | docker load
```
rfay: This isn't right, as the input to docker load is a stream of tar, so we don't want to just untar, we have to *uncompress* and pass that to `docker load`

rfay (changes requested):
Minor requests, will be great to have the macOS docker install more automated, but let's ask before we do it. Please test the docker load, I don't think it will work at all as is. 

I think we can just add 7z/7za for all 3 varieties of OS and we'll be fine. Can be done in this PR.

However, if you're sneaking the tar.xz decompression in, mention that in the PR.

Note that the tar.xz files are now in https://github.com/drud/ddev/releases/tag/v0.15.0

rfay:
~~I poked around for simple 7z or xz binaries to use on mac and linux and failed at that. Perhaps we can build and use, or steal from homebrew (for mac) but I was hoping for a sort-of-supported approach.~~

https://tukaani.org/xz/ has windows and a link to macOS package. The macOS package is from sourceforge of all places, it's a pkg; However, it seems only to be /usr/local/bin/xz, so I imagine we could ship it. 

All we really have to do is uncompress the tar.xz and then we're good to go, it doesn't actually have to be uncompressed to stdout.

rfay:
One note: The ddev_docker_images.tar.gz on v0.15.0 *was* corrupted. Hospital wifi failed on the upload :( I have updated it. But if *anybody* one sitting around in their package directory, it's broken, delete it, and the script will re-download it.

BrianGilbert:
xz binary is larger than 7z on mac, also the xz binary is kinda out of date, given 7z extracts xz files I think I'll go with it.  I think we can copy the built 7z binary from homebrew?

package_additions.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+# Create archive with docker additions for sprint (IDE and IRC).
+docker save briangilbert/cloud9-alpine:latest linuxserver/thelounge:latest | xz -z -9 > ${REPO_DIR}/docker_additions.tar.xz
```
rfay: Is the Docker build of this image available and checked in? If so, maybe add a comment with the link to it?

Could we change the name of this script to drupal_sprint_package_additions.sh?

package_additions.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
rfay, line 1: Recommend adding here:
```
set -o errexit
set -o pipefail
set -o nounset
```

These basically make all errors visible.

package_sprint.sh
```diff
@@ -10,9 +10,7 @@ STAGING_DIR=$STAGING_DIR_BASE/$STAGING_DIR_NAME
 FINAL_TARGET_DIR=/tmp
```
rfay: As suggested by @mradcliffe (I think) let's change this file to package_drupal_script.sh and update the README to match.

rfay (changes requested):
It's coming along great. Just running package_sprint.sh I get

```
++ docker save briangilbert/cloud9-alpine:latest linuxserver/thelounge:latest
++ xz -z -9
Error response from daemon: reference does not exist
+ cd /Users/rfay/tmp
+ tar -czf drupal_sprint_package.tar.gz drupal_sprint_package
+ zip -r -q drupal_sprint_package.zip drupal_sprint_package
++ ls '/tmp/drupal_sprint_package*'
ls: /tmp/drupal_sprint_package*: No such file or directory
+ printf '\033[32mThe sprint tarballs and zipballs are in .\033[0m\n'
The sprint tarballs and zipballs are in .
```

* The final statement about where the tarballs are could be more explicit :) That's just a related bug. Fixed in https://github.com/drud/quicksprint/pull/15
* We'll apparently need a `docker pull` of the `briangilbert/cloud9-alpine` before the docker save?
* Please don't use "latest" for the tag on cloud9-alpine, since it's always so ambiguous. Use a real tag that is expected to be predictable and not move.
* This leaves docker_additions.tar.xz laying around in the main directory where package_sprint.sh is run, seems that is expected to end up in $STAGING_DIR/ddev_tarballs right?.

BrianGilbert:
**Changes**
- Install docker if not present Mac OS
- Asks user if they need to install docker on windows
- Checks if docker present and links to installation instructions on linux
- Prompt to increase docker memory allocation (Mac OS)
- Core now compressed with xz compression as sprint.tar.xz
- Installation now copies core archive to ~/Sites/sprint 
- start_sprint script is also copied here
- start_sprint script extracts sprint-[yyyymmdd-hhmm] and updates multiple files with the timestamp value so that any given directory tells you exactly what you need to know
- 7za and sed are copied to /Sites/sprint/bin (windows only)
- Core is a subdir of sprint archive as drupal8 (this means that .ddev can be outside of the core dir and gives somewhere to house the start_clean scripts)
- start_clean will remove and resetup with checkout of head and create db during execution
- start_clean outputs url's of all the services at the end
- No longer package DB (set it up when using start_clean scripts instead)
- If present script called package_additions.sh used by package_sprint.sh during installation for additional containers
- probably other stuff I've missed..

package_additions.sh
BrianGilbert: added

BrianGilbert:
@rfay I think I've implemented all your feedback.

package_additions.sh
```diff
@@ -0,0 +1,8 @@
...
+set -o errexit
+set -o pipefail
+set -o nounset
+
+# Create archive with docker additions for sprint (IDE and IRC).
+docker pull briangilbert/cloud9-alpine:20180318 linuxserver/thelounge:94
```
mradcliffe: It looks like on docker-17.12.0-ce docker pull only supports exactly 1 argument.

> "docker pull" requires exactly 1 argument.
> See 'docker pull --help'
>
> Usage: docker pull [OPTIONS] NAME[:TAG|@DIGEST] [flags]
> 
> Pull an image or a repository fro a registry

mradcliffe (commented):
Ran into an issue using package script on a more recent version of docker (or an outdated version now?).

Edit: I'll fix it locally for myself so I can test a release on 2017 macbook and my slack box.

install_ddev.sh
```diff
@@ -14,26 +12,110 @@ BINOWNER=$(ls -ld /usr/local/bin | awk '{print $3}')
...
+#  -Copy required components to ~/Sites/sprint/
+#  -Do a git pull to ensure you have latest commits for core
+#  -Pre-load docker images for the sprint toolkit:
+#    -Drupal 8
+#    -phpmyadmin
+#    -Coud9 IDE
+#    -Thelounge IRC client
+#
+# Press y to continue
+# !!You don't need to hit enter!!.
```
mradcliffe, line 38: I think that people are going to hit Enter by default. I think this could be a follow-up to either accept Y/N to continue and repeat on any other key.

install_ddev.sh
```diff
@@ -14,26 +12,110 @@ BINOWNER=$(ls -ld /usr/local/bin | awk '{print $3}')
...
+        sleep 10
+        cp -rp /Volumes/Docker/Docker.app /Applications/
+        wait
+        open -a /Applications/Docker.app
+        hdiutil detach /Volumes/Docker
+
+        printf "
+        ${GREEN}
+        ####
+        # Please open Docker preferences and set Memory to 3.0 GiB on the Advanced tab.
```
mradcliffe, line 80: I like this step because it forces the user to remember to start Docker.

Steps necessary:

1. Open Docker.
2. Open Docker preferences 
3. Set memory (if supported by computer hardware)
4. Restart Docker
5. Wait for  Docker to start again
6. Press "y"

install_ddev.sh
```diff
@@ -14,26 +12,110 @@ BINOWNER=$(ls -ld /usr/local/bin | awk '{print $3}')
...
+        ${RESET}"
+        read -n1 DOCKMEM
+        if [[ ! $DOCKMEM =~ ^[Yy]$ ]]
+        then
+            exit 1
+        fi
+    fi
+
 elif [[ "$OS" == "Linux" ]]; then
     FILEBASE="ddev_linux"
```
mradcliffe, line 96: A follow-up task could be to try pyvenv and use pip to install (I usually do it in a Makefile). That could be automated although then the user would need to do it _again_ when going to ~/Sprints/whatever directory in the terminal.

package_drupal_script.sh
```diff
@@ -102,27 +102,35 @@ for dockerurl in $DOCKER_URLS; do
...
+if [ -f ${REPO_DIR}/package_additions.sh ]; then
+    # Package images for any additions.
+    source ${REPO_DIR}/package_additions.sh
+fi
+
 cd $STAGING_DIR_BASE
 tar -czf drupal_sprint_package.tar.gz $STAGING_DIR_NAME
 zip -r -q drupal_sprint_package.zip $STAGING_DIR_NAME
-printf "${GREEN}The sprint tarballs and zipballs are in $(ls $FINAL_TARGET_DIR/drupal_sprint_package*).${RESET}\n"
+printf "${GREEN}The sprint tarballs and zipballs are in $(ls $STAGING_DIR_BASE/drupal_sprint_package*).${RESET}\n"
```
mradcliffe: If things fail, let's say because a user was impatient and didn't wait for Docker to restart (*cough* me *cough*), then this message still prints out as a success even though it failed.

rfay: Hmm, this is not end-user - this is in the packaging script. Were you thinking about commenting elsewhere?

sprint/.ddev/docker-compose.sprint.yml
```diff
@@ -0,0 +1,43 @@
+version: "3"
+
+services:
+  ide:
+    container_name: ddev-${DDEV_SITENAME}-ide # This is the name of the container. It is recommended to follow the same name convention used in the main docker-compose.yml file.
+    image: briangilbert/cloud9-alpine:20180318
+    ports:
+    - 8000 # Coud9 is served from this port inside the container
+    environment:
```
mradcliffe, line 9: I was a bit confused at first because previously had been setting USERNAME and PASSWORD env vars to "drupal". Documentation follow-up to make sure those are aware it's the default username:password (pending Brian's issue of not having to specify it at all).

install_ddev.sh
```diff
@@ -49,10 +131,26 @@ else
...
     printf "${GREEN}Installed ddev bash completions in $bash_completion_dir${RESET}\n"
     rm /tmp/ddev_bash_completion.sh
 fi
 
-printf "${GREEN}ddev is now installed. Run \"ddev\" to verify your installation and see usage.${RESET}\n"
+mkdir -p ~/Sites/sprint
+cp start_sprint.sh ~/Sites/sprint/
+cp sprint.tar.xz ~/Sites/sprint/
+
+printf "
```
mradcliffe: Yes, sorry, this is where I mean to comment re: it printing out success regardless of if it failed because of my impatience.

sprint/Readme.txt
```diff
@@ -0,0 +1,27 @@
+To access
+Website:   http://sprint-[ts].ddev.local/
+           (U:admin P:admin)
+Mailhog:   http://sprint-[ts].ddev.local:8025/
+DB Admin:  http://sprint-[ts].ddev.local:8036/
+IRC:       http://sprint-[ts].ddev.local:9000/
+IDE:       http://sprint-[ts].ddev.local:8000/  
+           (U:username  P:password)
```
mradcliffe, line 8: It's right here. No doc follow-up needed. :-)

sprint/start_clean.sh
```diff
@@ -0,0 +1,56 @@
...
+    exit 1
+fi
+
+ddev remove
+ddev start
+wait
+ddev exec git fetch
+ddev exec git reset --hard origin/8.6.x
+ddev exec composer install
+ddev exec drush si standard --db-url=mysql://db:db@db:3306/db
```
mradcliffe: Follow-up for this to add `--account-pass`?

sprint/.ddev/docker-compose.sprint.yml
```diff
@@ -0,0 +1,43 @@
+version: "3"
```
rfay, line 1: Can you explain why this is added as a ddev service? It doesn't seem to me it has anything to do with ddev, and putting it in .ddev directory can cause a lot of confusion. If it's not a ddev service of some type can we put it in .quicksprint or something like that?

rfay (dismissed):
Hope to return and try out the actual running of things, but wanted to note my concern about putting the editor docker-compose into .ddev directory.

sprint/.ddev/docker-compose.sprint.yml
mradcliffe: Can it run through ddev-router without it being a ddev service? I think we were both independently following the docs about how to add containers so that we could start up the sprint toolset in one docker-compose command.

mradcliffe: Is there a more suitable way to do this?

package_additions.sh
BrianGilbert: have split into two commands now

sprint/.ddev/docker-compose.sprint.yml
BrianGilbert: the user/pass gets echoed out later, also had previously tried setting the env vars via the docker compose file and it didn't appear to be working

install_ddev.sh
BrianGilbert: have added a wait before the echo

sprint/start_clean.sh
BrianGilbert: Have added this now

sprint/.ddev/docker-compose.sprint.yml
BrianGilbert: As @mradcliffe said I was following ddev docs here..

install_ddev.sh
BrianGilbert: Have added not for user to ensure docker has restarted before continuing

package_drupal_script.sh
BrianGilbert: added a wait anyway.

sprint/.ddev/docker-compose.sprint.yml
BrianGilbert: @rfay is there a more suitable way that still allows them to get started at the same time as upping the drupal 8 site via ddev and keep the same url as the drupal 8 site?

rfay: The docs about adding services are for adding things like solr. Here we're adding a user-oriented web-based text editor right? Maybe I just don't understand the purpose, but I thought this was just a web-based editor, and is not integrated with ddev.

I may be lost, but if it's what I understand I'd bring up bring up the editor container with a completely separate and decoupled docker-compose command, and then bring up ddev.  I don't understand what they have to do with each other.

BrianGilbert: @rfay can we make use of the ddev-router if it's not so closely integrated?

For a Drupalcon first time sprinters workshop I specifically don't want it to have a different URL, as the user may have multiple instances running at once and it should be tied to the correct instance.

I just followed the ddev docs.. thats why it ended up this way..

install_ddev.sh
```diff
@@ -14,26 +12,111 @@ BINOWNER=$(ls -ld /usr/local/bin | awk '{print $3}')
...
-    printf "${RED}Sorry, this installer does not support your platform at this time.${RESET}\n"
+    printf "${RED}Sorry, this installer does not support your platform at this time.${RESET}\n"    
     exit 1
 fi
 
-if ! docker --version >/dev/null 2>&1; then
-    printf "${YELLOW}Docker is required for ddev. Download and install docker at https://www.docker.com/community-edition#/download before attempting to use ddev.${RESET}\n"
-fi
+echo "Installing docker images for ddev to use..."
+gzip -dc $(ls ddev_tarballs/ddev_docker_images*.tar.xz) | docker load
```
mradcliffe: I think I only commented in Slack on this, but this  and the similar line below also need a MacOS/Linux condition because gzip has native xz functionality in MacOS these days, but Linux does not. 

xzcat, similar to zcat, should extract xz to stdout.

`xzcat $(ls ddev_tarballs/ddev_docker_images*.tar.xz) | docker load`
`xzcat $(ls ddev_tarballs/docker_additions.tar.xz) | docker load`

mradcliffe (commented):
I should have made this comment not in Slack, but ran into something I probably fixed locally on a previous test.

install_ddev.sh
BrianGilbert: @mradcliffe do you know about it's existense and installation status on common dists?

mradcliffe: I have xzcat on Slackware 14.1/14.2 which is usually behind the times so I assume it's available on more hip distros like Ubuntu/Debian and Fedora/RHEL/CentOS.

rfay: Try `docker run -it debian xzcat -h`, `docker run -it ubuntu xzcat -h`, `docker run -it centos xzcat -h` etc.

mradcliffe:
Okay, this worked for me on Slackware more or less. I had an issue with theme files not loading probably due to permissions. This is probably not an issue on Debian/Ubuntu or RH-based distros, but I ended up with root:root owned files and containers.

BrianGilbert:
@mradcliffe was this perms issue in the guest or host?

Issue #7, opened 2018-03-07
Feature - generate database dump in packaging script


mradcliffe:
The database dump should be as up-to-date as possible with the latest Drupal core branch development so that there are no entity or database updates necessary. Core development in pre-release (pre-alpha) may have entity schema changes that are not stable or experimental.

Something like `ddev exec drush sql-dump --gzip --result-file=/some/path/to/dump.sql` or something? result-file probably needs to be the container volume, and then copied from there into .db_dumps?

rfay:
Good point, makes sense. Maybe the builder script should do a drush si and dump the result.

mradcliffe:
Closing. Brian implemented this.

PR #14, opened 2018-03-17
Generate database dump as part of package script

mradcliffe:
Solves #7


mradcliffe:
We need feedback on this approach as well: https://github.com/BrianGilbert/quicksprint/blob/master/sprint/start_clean.sh

rickmanelius:
Updated assignees. Moved to our project board for visibility.

mradcliffe:
Going to close this for now since the drupal_package script does this.

PR #17, opened 2018-03-30
Updated Docker installer versions, Fixed start_sprint windows script.

BrianGilbert:
-Add license info for sed.
-Updated docker installer versions
-Fixed start_sprint.cmd installer


rickmanelius (changes requested):
Looks reasonable from a code review. I haven't stepped through the actual start_sprint.cmd to confirm, so as long as someone else can thumbs up the actual end-user tet (I can test this weekend), then I would switch to a 👍

rickmanelius (approved):
On 2nd thought. Let's pull. The goal is to open this up to wider distribution to test and the big change from the scripts persective is Windows. Once this is pulled, a release can be generated to download the prepackaged container files for those testing.

PR #20, opened 2018-04-04
Increase compression used when creating package, remove definition of…

BrianGilbert:
… container versions, remove docker-compose.yaml, make sure text output in echoed bits is on newline.


rfay (approved):
All my comments were resolved in slack (I was doing a review and suddenly everything was already fixed!). 

One test user today though said they couldn't get their site to start without removing the docker-compose-sprint.yml, so you might want to give that a look.

BrianGilbert:
@rfay What OS were they on, I tested yesterday and it didn't cause any issues here.

Issue #31, opened 2018-04-05
Windows and mac docker versions are not current


rfay:
Both Windows and Mac docker_installs provide docker 17.09, not the current recommended 18.03. Please update to the current version. As noted elsewhere, 
* please name the install to contain the version number
* Remove any spaces in the names (makes scripting far better and easier) 

The best place to get the docker-for-mac and docker-for-windows releases is on their release notes pages, for example https://docs.docker.com/docker-for-mac/release-notes/

BrianGilbert:
I'm using the most current versions linked on the Docker website 
```
DOCKER_URLS="https://download.docker.com/mac/stable/23751/Docker.dmg https://download.docker.com/win/stable/16762/Docker%20for%20Windows%20Installer.exe"
```
https://docs.docker.com/docker-for-mac/release-notes/
https://docs.docker.com/docker-for-windows/release-notes/


Addressing the dot points though..

rfay:
Yeah, 23751 (mac) is correct to download, but it not what was in the package in v0.0.5.

I do think that if the packaging script were run against an existing directory, it would have respected old docker stuff that was there, but I don't think @rickmanelius  had ever done that before last night.

rfay:
The key point is: Build it and make sure it has the right stuff inside.

rfay:
Also, the script shouldn't be installing docker. Make people do that, and start docker. It will never turn out right, I don't think, if the script tries to do it.

BrianGilbert:
I know it does as the version of Docker I have installed locally was done by the script.. and it's 18.03.0-ce-mac60 (23751)

BrianGilbert:
@Rfay, perhaps you should look at the open PR :)

rfay:
I confirm that this is fixed for both macOS and windows in #32

BrianGilbert:
Suspect this was a packaging error

Issue #30, opened 2018-04-05
install_ddev.cmd fails to run docker install (and doesn't). Must run docker, not just install.


rfay:
Multiple problems with the install_ddev.cmd and the docker version:

* The install_ddev.cmd fails to find the right docker version (see screenshot)
* The actual provided docker-ce version is 17.09.0, we should be shipping 18.03.0 (#31 - problem on both windows and mac)
* docker must be running for the next step (the ddev installation fails without it). So just installing docker (on windows) is not enough - it must be run.

![failure_to_install_docker](https://user-images.githubusercontent.com/112444/38388353-42afc6b0-38d8-11e8-9b74-d8a4f922fb2d.png)

BrianGilbert:
@rickmanelius had you previously run the packaging script and not deleted the stuff in tmp before updating it yesterday?

rfay:
This was on a windows machine; the packaging script can't be run on Windows, so I don't think that's a problem. The script having trouble here is install_ddev.cmd - on WIndows.

BrianGilbert:
@rfay yes, but if Rick did what I asked when he created the release yesterday it would have kept the old version of docker installers that had been downloaded, causing the version discrepencies

rfay:
You were probably asking the question in the wrong issue I guess @BrianGilbert  - This issue is about failures of the windows install script.

BrianGilbert:
@rfay yes, I'm aware.. but the provided version of docker is based on what happened when the package script was run.

BrianGilbert:
Docker installation removed, user needs to manually install it now, packaging script has option to include installed when building artifact.

Issue #28, opened 2018-04-05
drupal_sprint_package.tar.gz/.zip should have the release version in the filename.


rfay:
When you download a sprint package, it's important to be able to know what release version is in it. So we should be creating drupal_sprint_package.v0.0.5.tar.gz and .zip, for example.

BrianGilbert:
done

Issue #26, opened 2018-04-05
Scripts need error checking


rfay:
All of the user-oriented scripts (start_ddev, start_sprint, start_clean) need error checking so they don't continue on when an error is encountered.

```
set -o errexit
set -o pipefail
set -o nounset
```

Maybe more.

rfay:
For example. install_ddev.sh with docker not running results in:

```
Installing docker images for ddev to use...
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
tar: /tmp: Not found in archive
tar: -C: Not found in archive
tar: ddev_tarballs/ddev_macos.v0.16.0.tar.gz: Not found in archive
tar: Error exit delayed from previous errors.
chmod: /tmp/ddev: No such file or directory
Ready to place ddev in your /usr/local/bin.
mv: rename /tmp/ddev to /usr/local/bin/ddev: No such file or directory
/usr/local/bin/brew
cp: /tmp/ddev_bash_completion.sh: No such file or directory
Installed ddev bash completions in /usr/local/etc/bash_completion.d
rm: /tmp/ddev_bash_completion.sh: No such file or directory


######
#
# Your ddev and the sprint kit are now ready to use,
# execute the following commands now to start:
#
# cd ~/Sites/sprint
# ./start_sprint.sh
#
######
```

BrianGilbert:
Now checks that docker is runnnig or exits with message.

PR #32, opened 2018-04-06
Fixes for v0.0.6 release

BrianGilbert:
[#21] Ensure ddev is added to path, Let the user know we exited if they didn’t hit Y/y, add that install should be run via terminal on mac and Linux.
[#26] Added error checking to all shell scripts.
[#28] add release version to archive filename.
[#30] Don’t install docker by default, improve messaging about Docker memory requirement.
[#31] Add version number to Docker installers, remove spaces from filenames, make packaging of installers optional. Don’t automatically install docker on mac, make the user do it manually instead. Improve visibility of requirement to increase docker memory limit.
[#30 #31] When running packaging script, delete and build it clean if staging dir already exists.

### DIFF
--- a/package_drupal_script.sh
+++ b/package_drupal_script.sh
@@ -31,8 +31,14 @@ BINOWNER=$(ls -ld /usr/local/bin | awk '{print $3}')
 USER=$(whoami)
 
 if [ -d "$STAGING_DIR" ] && [ ! -z "$(ls -A "$STAGING_DIR")" ] ; then
-    echo -n "The staging directory already has files. Deleting them and recreating everything."
+    printf "${RED}The staging directory already has files. Deleting them and recreating everything.${RESET}"
     rm -rf $STAGING_DIR
+    if [ -e $STAGING_DIR_BASE/drupal_sprint_package$QUICKSPRINT_RELEASE.tar.gz ] ; then
+        rm $STAGING_DIR_BASE/drupal_sprint_package$QUICKSPRINT_RELEASE.tar.gz
+    fi
+    if [ -e $STAGING_DIR_BASE/drupal_sprint_package$QUICKSPRINT_RELEASE.zip ]; then
+        rm $STAGING_DIR_BASE/drupal_sprint_package$QUICKSPRINT_RELEASE.zip
+    fi
 fi
 
 SHACMD=""
@@ -64,6 +70,8 @@ if ! docker --version >/dev/null 2>&1; then
     printf "${YELLOW}Docker is required to use this package. Download and install docker at https://www.docker.com/community-edition#/download before attempting to use ddev.${RESET}\n"
 fi
 
+cd $STAGING_DIR
+
 printf "
 ${GREEN}
 ####
@@ -72,24 +80,25 @@ ${GREEN}
 ####
 ${RESET}"
 read -n1 INSTALL
-if [[ ! $INSTALL =~ ^[Yy]$ ]]
+if [[ $INSTALL =~ ^[Yy]$ ]] ; then
     # Download current docker installs
+    printf "
+${GREEN}
+####
+# Downloading docker installers.
+###
+${RESET}"
     mkdir -p docker_installs
     for dockerurl in $DOCKER_URLS; do
         fname=$(basename $dockerurl)
-        if [ ! -f "docker_installs/$fname" ] ; then
-            if [[ $fname = *"dmg"* ]]; then
-                curl -sSL -o "docker_installs/Docker-$DOCKER_VERSION_MAC.dmg" $dockerurl
-            elif [[ $fname = *"exe"* ]]; then
-                curl -sSL -o "docker_installs/Docker-$DOCKER_VERSION_WIN.exe" $dockerurl
-            fi
+        if [[ $fname = *"dmg"* ]]; then
+            curl -sSL -o "docker_installs/Docker-$DOCKER_VERSION_MAC.dmg" $dockerurl
+        elif [[ $fname = *"exe"* ]] ; then
+            curl -sSL -o "docker_installs/Docker-$DOCKER_VERSION_WIN.exe" $dockerurl
         fi
     done
-then
-    exit 1
 fi
 
-cd $STAGING_DIR
 mkdir -p ddev_tarballs
 TARBALL="ddev_docker_images.$LATEST_VERSION.tar.xz"
 SHAFILE="$TARBALL.sha256.txt"
--- a/package_drupal_script.sh
+++ b/package_drupal_script.sh
@@ -15,7 +15,7 @@ STAGING_DIR_NAME=drupal_sprint_package
 STAGING_DIR_BASE=~/tmp
 STAGING_DIR=$STAGING_DIR_BASE/$STAGING_DIR_NAME
 REPO_DIR=$PWD
-QUICKSPRINT_RELEASE=v0.0.6
+QUICKSPRINT_RELEASE=$(git describe --tags --always --dirty)
 
 # The version lines on the following few lines need to get changed any time the url are changed on the line below.
 DOCKER_URLS="https://download.docker.com/mac/stable/23751/Docker.dmg https://download.docker.com/win/stable/16762/Docker%20for%20Windows%20Installer.exe"
@@ -29,6 +29,9 @@ RESET='\033[0m'
 OS=$(uname)
 BINOWNER=$(ls -ld /usr/local/bin | awk '{print $3}')
 USER=$(whoami)
+
+# Ensure XZ is installed
+command -v xz >/dev/null 2>&1 || { echo >&2 "${RED}I require xz command but it's not installed. Aborting.${RESET}"; exit 1; }
 
 if [ -d "$STAGING_DIR" ] && [ ! -z "$(ls -A "$STAGING_DIR")" ] ; then
     printf "${RED}The staging directory already has files. Deleting them and recreating everything.${RESET}"
@@ -76,6 +79,8 @@ printf "
 ${GREEN}
 ####
 # Shall we package docker installers for mac and windows with the archive?
+#
+# Press y to include installers, or any other key to continue.
 # !!You don't need to hit enter!!.
 ####
 ${RESET}"
@@ -165,4 +170,4 @@ cd $STAGING_DIR_BASE
 tar -czf drupal_sprint_package$QUICKSPRINT_RELEASE.tar.gz $STAGING_DIR_NAME
 zip -9 -r -q drupal_sprint_package$QUICKSPRINT_RELEASE.zip $STAGING_DIR_NAME
 wait
-printf "${GREEN}The sprint tarballs and zipballs are in $(ls $STAGING_DIR_BASE/drupal_sprint_package*).${RESET}\n"
+printf "${GREEN}The sprint tarballs and zipballs are in $STAGING_DIR_BASE.${RESET}\n"
--- a/package_drupal_script.sh
+++ b/package_drupal_script.sh
@@ -30,13 +30,8 @@ BINOWNER=$(ls -ld /usr/local/bin | awk '{print $3}')
 USER=$(whoami)
 
 if [ -d "$STAGING_DIR" ] && [ ! -z "$(ls -A "$STAGING_DIR")" ] ; then
-    echo -n "The staging directory already has files. Do you want to continue (y/n)? "
-    read answer
-    if echo "$answer" | grep -iq "^y"; then
-        echo "Continuing with downloads, existing files will be respected, mostly."
-    else
-        exit 1
-    fi
+    echo -n "The staging directory already has files. Deleting them and recreating everything."
+    rm -rf $STAGING_DIR
 fi
 
 SHACMD=""
--- a/package_drupal_script.sh
+++ b/package_drupal_script.sh
@@ -170,4 +170,20 @@ cd $STAGING_DIR_BASE
 tar -czf drupal_sprint_package$QUICKSPRINT_RELEASE.tar.gz $STAGING_DIR_NAME
 zip -9 -r -q drupal_sprint_package$QUICKSPRINT_RELEASE.zip $STAGING_DIR_NAME
 wait
-printf "${GREEN}The sprint tarballs and zipballs are in $STAGING_DIR_BASE.${RESET}\n"
+
+printf "
+${GREEN}
+####
+# The built sprint tarballs and zipballs are now in ${YELLOW}$STAGING_DIR_BASE${GREEN}.
+#
+# Now deleting the staging directory.
+####
+${RESET}"
+rm -rf $STAGING_DIR_NAME
+wait
+printf "
+${GREEN}
+####
+# Finished
+####
+${RESET}"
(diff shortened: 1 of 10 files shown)


rfay:
```
####
# Shall we package docker installers for mac and windows with the archive?
# !!You don't need to hit enter!!.
####
```
I apparently *do* need to hit something :) Maybe Y/n prompt? What's the intent for a normal build? No docker?  

Hitting "y" worked for me the first time. The second time I did it (after validating the downloaded versions) I didn't want to bother, so hit "n", and nothing happened. Despite the instructions I also hit "<enter>" and also hit `<n><\n>`. Turned out that *one of those* worked, but it gave no feedback at all so I was really wondering, Eventually it showed download results while I was writing this.

rfay:
`/Users/rfay/go/src/github.com/drud/quicksprint/package_additions.sh: line 9: xz: command not found`

I'm doing a `brew install xz` now, and very few users will ever run this script. However, it might want to test for the existence if necessary tools at the beginning instead of at the end. `command -v` might do well, https://stackoverflow.com/a/677212/215713

rfay:
At the end of the script it displays everything in the drupal_sprint_package staging dir, which doesn't seem very useful. It's a result of `printf "${GREEN}The sprint tarballs and zipballs are in $(ls $STAGING_DIR_BASE/drupal_sprint_package*).${RESET}\n"` - How about we change that ls to something more specific so it will only show the tarball and zipball?

Also, I note that v0.0.6 seems to be wired into the QUICKSPRINT_RELEASE. It would be loads better to get it from git. Let's make QUICKSPRINT_RELEASE=$(git describe --tags --always --dirty) - that will get the current tag in there, or if it's not on a tag, something like the nice tag of this PR currently, v0.0.5-12-g70fbcaf

```
The sprint tarballs and zipballs are in /Users/rfay/tmp/drupal_sprint_packagev0.0.6.tar.gz
/Users/rfay/tmp/drupal_sprint_packagev0.0.6.zip

/Users/rfay/tmp/drupal_sprint_package:
COPYING
SPRINTUSER_README.md
bin
ddev_tarballs
install_ddev.cmd
install_ddev.sh
licenses
sprint.tar.xz
start_sprint.cmd
start_sprint.sh.
```

package_drupal_script.sh
```diff
@@ -15,8 +15,12 @@ STAGING_DIR_NAME=drupal_sprint_package
 STAGING_DIR_BASE=~/tmp
 STAGING_DIR=$STAGING_DIR_BASE/$STAGING_DIR_NAME
 REPO_DIR=$PWD
+QUICKSPRINT_RELEASE=v0.0.6
```
rfay: Please use `QUICKSPRINT_RELEASE=$(git describe --tags --always --dirty)` instead of wiring to something that has to be changed all the time.

package_drupal_script.sh
```diff
@@ -138,7 +162,7 @@ if [ -f ${REPO_DIR}/package_additions.sh ]; then
 fi
 
 cd $STAGING_DIR_BASE
-tar -czf drupal_sprint_package.tar.gz $STAGING_DIR_NAME
-zip -9 -r -q drupal_sprint_package.zip $STAGING_DIR_NAME
+tar -czf drupal_sprint_package$QUICKSPRINT_RELEASE.tar.gz $STAGING_DIR_NAME
+zip -9 -r -q drupal_sprint_package$QUICKSPRINT_RELEASE.zip $STAGING_DIR_NAME
 wait
 printf "${GREEN}The sprint tarballs and zipballs are in $(ls $STAGING_DIR_BASE/drupal_sprint_package*).${RESET}\n"
```
rfay: We don't need to show the whole directory, let's just show the tar/zipballs.

rfay:
```
STAGING_DIR_NAME=drupal_sprint_package
STAGING_DIR_BASE=~/tmp
STAGING_DIR=$STAGING_DIR_BASE/$STAGING_DIR_NAME
```

These are all currently set up so the staging directory and the contents of the tarball are the same, so if I untar the sprint package to test it, it overwrites the staging directory (or perhaps would have things in there it shouldn't). Could we 1) Change the name of the tarball to something else? or 2) Change the name of the staging directory? Maybe both?

rfay:
Let's make the xz unpack of the drupal repo quiet (in start_sprint.sh/cmd) instead of showing all that huge extraction of the .git directory and everything. That's pretty foreign to most people.

rfay:
start_clean.sh doesn't start the ddev site. Shouldn't it? At the end of the the script it errors? out with `Project is not currently running. Try 'ddev start'.`

I think it doesn't work.

I'm not sure the "clean" idea is meaningful. What is "clean" supposed to be? Blow away everything? 

I *am* able to `ddev start` and get to the install screen. However, it doesn't have a default db loaded, probably you want them to go through the install process?

rfay (commented):
This is much better, IMO you can pull when ready. I made several comments and some code comments.

package_drupal_script.sh
BrianGilbert: @rfay not sure what you mean here?

rfay:
Oh but I have only tested on Mac so far. I will try windows and Linux this weekend.

rickmanelius:
@rfay So based on your last comments... pull this and then submit additional PRs after? I would rather adopt that approach before my next round of testing. I'm on a fiber connection, so I can push a release for others to start working against for next round.

BrianGilbert:
@rickmanelius that will make it hard to track these comments, unless new issues are created

rickmanelius:
@BrianGilbert Fair enough. I'll work against this branch and rebuild/test locally.

package_drupal_script.sh
BrianGilbert: done

rickmanelius:
# Testing

Performed by Rick Manelius as of 2018-04-06

## macOS

### Hardware Setup

I’m using a laptop with the following high-level specifications:

* MacBook Pro: Retina 15-inch Mid 2015
* 2.2 Ghz Intel Core i7
* 16 GB 1600 MHz DDR3 RAM
* 250 GB Hard Drive

### Software Setup

* macOS Sierra 10.12.4
* Xcode
* Git 2.14.3
* Note that I specifically removed all traces of `go`, `ddev`

### Preparation


#### Cleanup

* `brew uninstall go` && `brew uninstall ddev`
* Remove "Docker for Mac" from menu and remove application from /Applications
  * Note last time I had some files left over in /user/local/bin. This time this was not present.

#### Artifacts

* Checkout
  * `git clone git@github.com:drud/quicksprint.git`
  * `git fetch --all`
  * `git checkout 6c13e62`
* Building via MAINTAINER_README.md
  * I selected "yes" to include the installers.
  * I'm seeing the creation of ~/tmp
  * I'm seeing the creation of ddev binary zips into ~/tmp
  * I'm seeing the docker installers in ~/tmp but I didn't see output to indicate that it happend.
  * Having removed "Docker for Mac" completely, I see that package_additions.sh failed.
    * This is fixed once Docker for Mac was installed.
  * I'm seeing the docker extra docker images pull.
  * I see the success message.

#### Using as End-user

* SPRINTUSER_README.md
* ./install_ddev.sh
  * I'm seeing the containers download
  * I can verify ddev binary is installed
  * `cd ~/Sites/sprint`
  * `cd sprint-20180406-1515`
  * I ran ./start_clean.sh
    * I noted that I was asked to run `ddev start`.
    * Since running `ddev start` does bring me up to an active site, I could infact ignore ./start_clean.sh
  * I re-ran ./start_clean.sh and now I see all the other containers.
    * ./start_clean.sh clean may wish to have the project in a `ddev start` first and then execute to make this consistent.
  * SUCCESS!
  * We should probably specify admin/admin as the user/password.
    * Or we could just run a final `ddev exec drush user-login`

### Recommendations

* Perhaps inform the user that the Docker for X installers were stored in ~/tmp.  
* Perhaps check for Docker when packaging https://github.com/drud/quicksprint/pull/32#issuecomment-379372684
* Perhaps note when the zip is being generated.
* * ./start_clean.sh clean may wish to have the project in a `ddev start` first and then execute to make this consistent.

rickmanelius:
Open items/questions:

* [x] Docker requirement for building package. I got ahead of myself to test as an end-user by removing Docker for Mac. This caused the initial run of ./package_drupal_script.sh to end with a message that "/Users/rmanelius/git/quicksprint/package_additions.sh: line 7: docker: command not found". See https://github.com/drud/quicksprint/blob/6c13e62a80c675f4a637332746de2c9431f872bc/package_drupal_script.sh#L171. We should probably just not this in the MAINTAINER_README.md as a dependency to do this right.
* [x] There is a long pause after Status: Downloaded newer image for linuxserver/thelounge:94 is run and the tar is being generated https://github.com/drud/quicksprint/blob/master/package_drupal_script.sh#L140. I would recommend a quick output to state what's happening versus it's frozen.

rickmanelius:
Testing so far, I can now get the zips generated. Moving on to testing as a macOS user. About to test as an end-user after I remove Docker for Mac and ddev. I am noting the one dependency of having docker installed to make the build, but that shouldn't trip too many people up.

rickmanelius:
Finished and updated my review for macOS. Overall, the language and guidance between steps is much better then my previous pass https://github.com/drud/quicksprint/issues/21#issuecomment-378830583. I was also successful, which is great! I believe a few people have mentioned testing Linux/Windows to confirm the ability to get to a success.

FWIW, I put a few other recommendations for either messaging (when something is happening that takes a while) or potentially re-ordering some things (such as needing to run `ddev start` before running the cleanup script). That said, the most important thing is ensuring we have cross OS success, and other nice upgrades are just icing on the cake.

BrianGilbert:
@rickmanelius I added a check that docker is running in to the relevant scripts in https://github.com/drud/quicksprint/pull/32/commits/3be763e59e1d7128fea986b418b30abfeec16816 & https://github.com/drud/quicksprint/pull/32/commits/91e5690c22c078cad34176622adbedaa4aada88c, which I think I committed after you had pulled to test

geerlingguy:
Testing this PR on Windows 10 Pro today... as soon as my kids will take a nap :P — I'll update here with any notes:

  - In the first step in the maintainer readme, it says to run the packaging script... but this doesn't seem to be able to be done on Windows at all? Maybe we should add a note that the _packaging_ can only be done on Mac/Linux.
  - Unzipping the 21,000+ items in the drupal_sprint_package.zip file on my Windows 10 laptop (SATA III SSD, Lenovo T420 with a Core i5 processor) took more than 25 minutes 😮. Is there any way to cut down on the volume of tiny files that need to be expanded, to speed up this operation? I can imagine every person in the sprint room would need to sit and do nothing for about the first half hour if this is common on an average Windows machine.
  - After Docker CE was installed, and after running the install_drud script (which had no errors), and after a reboot to finish Docker installation, I ran the `start_sprint.cmd` command inside my user's ~/Sites/sprint folder, and it opened a cmd window, did something super quick, but then closed the window and I didn't see any difference in Docker or anywhere else. Might it be failing silently? I can do a screencast or something to show you what I mean.
    - Running `docker ps` didn't show any running containers.
    - It looks like the script may silently fail on the 'running as administrator' permissions check; **we should highlight that you have to right-click the file and select 'Run as Administrator' otherwise it will fail**.
    - I added a few `pause`s into the start cmd and what's really weird as that when I ran it as Administrator, it would bail out at the step where it started expanding the `sprint.tar.xz` file, but when I ran it _not_ as Administrator, it started extracting the files and seemed to succeed... very strange.
  - Once I got through the `start_sprint.cmd` stuff above, I opened the sprint folder it created, then ran the `start_clean.cmd` file.
    - I saw a lot of stuff like `[][0m` at the start of lines (and `[][33m`). This was in PowerShell.
    - At this point it started pulling a bunch of Docker containers (I thought these were already downloaded with the rest of the archive? Maybe not).
    - It took maybe 5 min or so to bring up all the containers, but I got a bunch of errors about drives not being shared because I didn't see the windows admin prompt to allow sharing directories with Docker. I accepted that prompt after the fact, and also entered my password to allow vpnkit.exe to do stuff, and the ddev start command seemed to finish at some point.
    - I visited http://sprint-20180704-1506.ddev.local/ at this point but it still wouldn't load.
    - I ran `ddev start` again, and then noticed that it mentioned 'you must manually add the following entry to your hosts file' (though there weren't any directions in the sprintuser readme on how to do it...).
    - I edited my hosts file (`C:\Windows\System32\drivers\etc\hosts`) as Administrator, adding the entry that was recommended.
    - VOILA! I got the Drupal installer, running 8.6.0-dev
  - Drupal Installer:
    - I had to enter the database username/password/database/host manually, and found that the hostname is `ddev-sprint-20180704-1506-db`, but couldn't find what the default user/pass/db are. Looking for that now.
    - I ran the `start_clean.cmd` command again to see if it would fix my problems... but it seemed to sit and wait after a few steps, and after waiting 5 or so minutes with no output, I just canceled it out and ran `ddev start` again. That seemed to go through without any errors and got me back to my command prompt.
    - I searched around the internet, and eventually found there's a `ddev describe` command that can give things like database credentials, etc. So I ran that, found the defaults are `db`/`db`/`db` with the hostname listed earlier. Drupal installed, yay!

screen shot 2018-04-07 at 3 27 25 pm

BrianGilbert:
@geerlingguy something went wrong if you had that many files.. it should only be 31 files in 5 folders when extracting the drupal_sprint_package.zip file.

geerlingguy:
@BrianGilbert - I downloaded the release as instructed from https://github.com/drud/quicksprint/releases (drupal_sprint_package.zip). It was 1.34 GB and seemed to have all those files inside once expanded...

BrianGilbert:
@geerlingguy There are heaps of fixes in this PR that aren't in the latest downloadable release. Also the last release was packaged by @rickmanelius and the package archives seem to have some weird issues/corruptions.. I've removed them now.

geerlingguy:
@BrianGilbert - Thanks; hopefully we can get this merged ASAP and get a 0.0.6 so the release packages are updated :)

I've completed the rest of my notes from the Windows 10 install process. If I can, I'll try doing it again sometime before I head out for DrupalCon if there's a newer release (no way I lug the Thinkpad with me all the way to Nashville :P ).

BrianGilbert:
@geerlingguy I think I've addressed all of your points.. either already in this PR, or just now as addirtions to this PR.

PS: You don't need to know the DB details if using PHPMyAdmin thats packaged with ddev.

geerlingguy:
@BrianGilbert - Is the start_clean.cmd supposed to also end up installing Drupal? I needed the details because when I loaded the sprint project URL it took me to the Drupal installer.

BrianGilbert:
@geerlingguy the start_clean.cmd script was missing a few steps, your feedback pointed me there so thanks.. I've added them in now..

geerlingguy:
@BrianGilbert - Awesome! I'm not leaving until Monday around noon, so if there's a new release between now and then I can download to the Windows machine based on this PR's work, I'll re-test and verify things are working more reliably :)

BrianGilbert:
@geerlingguy I'm going to make one now.. hopefully I can upload the artifacts without it timing out this time.. (crappy australian broadband)

PR #35, opened 2018-04-08
Fix if statement so start_clean.sh script continues, fixes #34

### LINKED ISSUE #34: start_clean.sh exits instead of continuing.

BrianGilbert:


rfay (approved):
This seems fine if it's been tested. I only had a problem with the `ddev remove` line, and the fix here is the same fix I made. So I have only tested line 51.

PR #37, opened 2018-04-08
Feature/34 share drives windows

BrianGilbert:
Also remove non-working checks that docker was running (changed to text prompts instead)


rfay (approved):
This seems fine.

Issue #38, opened 2018-04-08
Prompt user to enter real name field in The Lounge


BrianGilbert:

BrianGilbert:
changed pre-filled text to something that prompts user to enter something

PR #39, opened 2018-04-08
[#38] Prompt the userr to changethe name field rather than leaving it…

BrianGilbert:
… as default.


rfay (approved):
Seems reasonable, should be tried if it hasn't been.

PR #46, opened 2018-04-09
Make windows scripts work 

rfay:
There were problems throughout the windows scripts; this is a first attempt to get them working.  It will need significant testing.

### DIFF
--- a/install_ddev.cmd
+++ b/install_ddev.cmd
@@ -27,8 +27,8 @@ PAUSE
 
 cd %~dp0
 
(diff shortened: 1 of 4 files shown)


BrianGilbert:
where docker.exe 
may be a simpler check (easier for people to read the code and understand)
https://ss64.com/nt/where.html

rfay:
WHERE docker and WHERE docker.exe have been the same thing for at least a decade. Back in the day there would have been a possibility for "docker.com" as well, but that's long since gone. Doesn't matter.

I may add another commit to this PR.

PR #47, opened 2018-04-09
[#44] Changed Thelounge port to 8999, fixes #44

### LINKED ISSUE #44: Move theLounge to a less-likely port
Port 9000 is likely to have lots of conflicts, so not a good port for theLounge; How about port 9999?

Both xdebug and php-fpm both use port 9000 by default

BrianGilbert:
Closes #44

### DIFF
--- a/sprint/.ddev/.thelounge/config.js
+++ b/sprint/.ddev/.thelounge/config.js
@@ -27,7 +27,7 @@ module.exports = {
 	// @type     int
 	// @default  9000
 	//
-	port: 9000,
+	port: 9999,
 
 	//
 	// Set the local IP to bind to for outgoing connections. Leave to undefined
--- a/sprint/.ddev/.thelounge/config.js
+++ b/sprint/.ddev/.thelounge/config.js
@@ -27,7 +27,7 @@ module.exports = {
 	// @type     int
 	// @default  9000
 	//
-	port: 9999,
+	port: 8999,
 
 	//
 	// Set the local IP to bind to for outgoing connections. Leave to undefined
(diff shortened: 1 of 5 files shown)


package_drupal_script.sh
```diff
@@ -6,7 +6,7 @@ set -o nounset
 
 # Maximise compression
 export XZ_OPT=-9e
-export GZIP=-9e
+export GZIP=-9
```
mradcliffe: This change worked for me on linux.

rfay: I don't understand why gzip settings are being changed in this PR. Isn't this about a port change? It's best to stick to one thing.

rfay: This made it in in https://github.com/drud/quicksprint/pull/51

rfay:
This looks fine to me but I haven't manually tested it yet.  Should be pretty easy.

rfay:
Just needs rebase to have circle tests and I'll review in the morning. If you authorize me, I'll also pull while you're sleeping.

BrianGilbert:
@rfay happy with you merging when your ready

rfay:
This doesn't work at all - we get a 502. The problem is that the internal port is still 9000, so the docker-compose.sprint.yaml should be

```
    ports:
    - 9000 # Thelounge is served from this port inside the container
```

Please test your changes with a full run-through.

BrianGilbert:
@rfay changed port to 8999 so they are all in 8k range, tested locally and confirmed working.

BrianGilbert:
@rfay Circle fail looks like a network error trying to download stuff to run the test, I don't have rights to retrigger the build.

rfay:
Actually it's not that you don't have the right, it's that circleci is screwy on this exact thing. There's just no appropriate link for it. If you see the "rebuild with ssh" button, it's ok to use that too.

I hit it.

Another technique is to `git commit --amend` and make a change to the description and then `git push --force-with-lease`.

rfay (approved):
This works on Windows and macOS, should be good to go. Thanks!

Issue #40, opened 2018-04-08
Windows 10 Pro sprint user installation testing - second attempt


geerlingguy:
After 0.0.6 (plus 3 commits, at `c1b1ebe`), I ran the `./package_drupal_script.sh` script on my Mac, then copied the resulting `.zip` file over to my PC (it was ~600 MB; I didn't include the Docker installers).

I then did the following:

  1. I reset my local environment after yesterday's tests:
    1. Tried deleting my old `drupal_sprint_package` directory, but it seems the directory `C:\Users\jgeerling\Downloads\drupal_sprint_package\drupal_sprint_package\drupal8\core\modules\content_moderation\tests\src\Kernel` screwed something up. I am the owner, but apparently have no permission to delete, modify, etc. the file (and thus the entire file path is blocked from being renamed/deleted/etc.). It looks like I may have to boot in safe mode to be able to delete that folder :(
    1. I didn't uninstall ddev after yesterday's testing... how should I do that?
    1. I deleted the contents of `~/Sites/sprint`.
  1. I expanded the `.zip` file (was much quicker this time around!).
  1. I opened the new `drupal_sprint_package` directory.
  1. I ran `install_ddev.cmd`, pressed `y`, and the script did whatever it needed to do pretty quickly.
  1. At this point, I didn't see anything inside `~/Sites/sprint`, -so I'm guessing I'll need to figure out how to uninstall ddev so the `install_ddev.cmd` command does something?- (see notes in later comments).
  1. After patching the `install_ddev.cmd` file's Docker check as mentioned below and running it again, the `~/Sites/sprint` directory was populated.
  1. I opened up the `~/Sites/sprint` directory and ran the `start_sprint.cmd` file, and it opened and closed very quickly 🤷‍♂️ 
    1. There was no error message about it, but it looked like I had to run it as administrator. So I ran it as administrator, and it still opened and closed very quickly 🤷‍♂️ 

It seems like the script exits out when it hits this part:

```
REM # This script creates a new drupal 8 instance in the current directory ready to sprint on an issue.
REM #Create a timestamp
set HOUR=%time:~0,2%
set SANEHOUR=%HOUR: 0=24%
set SANEHOUR=%HOUR: =0%
set TIMESTAMP=%date:~10,4%%date:~7,2%%date:~4,2%-%SANEHOUR%%time:~3,2%
```

[...]

PR #48, opened 2018-04-09
[#40] So that user will know what happened, pause before exiting in case user ran script from explorer

BrianGilbert:


mradcliffe:
This works for me on MacOS.

rfay:
This needs a rebase.

rfay:
I removed the administrator stuff from start_sprint.cmd in https://github.com/drud/quicksprint/pull/46, don't think it will be relevant. We'll reopen this if it turns out to be important.

PR #50, opened 2018-04-10
Add circleci build and test

rfay:
This adds a circleci build; it will only work on Linux, but it's something, and it gives manual testers an explicit version to build against.


mradcliffe:
I'm not a CircleCI expert, but the commands seem to make sense.

Follow-up could be to split the packages into windows/macos/linux builds if we need to upload directly to github, but that probably doesn't matter if we're using s3.

rfay:
Thanks for looking @mradcliffe - Right now this is mostly about making the builds repeatable and less onerous (and doing a wee bit of testing). We don't have an easy way to test on windows or macOS unfortunately; there's not a good platform for either that I have set up. Even though Circle will do macOS I haven't been able to get docker going on it. And even though appveyor does a nice job with Windows, its docker won't do linux images, only Windows images.

BrianGilbert:
As per @mradcliffe 's comments no Circle expert either, but I can see whats happening here and it looks sensible to me too. 👍

rfay:
Pulled it based on informal approval, thanks both!

PR #56, opened 2018-04-10
Fixes #40: Add details to final Windows user instructions in SPRINTUSER_README.md.

### LINKED ISSUE #40: Windows 10 Pro sprint user installation testing - second attempt

geerlingguy:


SPRINTUSER_README.md
```diff
@@ -17,6 +17,9 @@ Installation and Startup:
 2. Start Docker (it must be running to install ddev)
 3. Open Docker preferences and set docker memory allocation to 3.0GB in the Advanced section (required for container to be started)
 3. Install DDEV-Local. (macOS/Linux): Run install_ddev.sh via terminal from the directory where this SPRINTUSER_README.md is with `./install_ddev.sh`; (Windows): run `install_ddev.cmd`.
-4. Create Drupal 8 instance to use during sprint: (macOS/Linux): cd to `~/Sites/sprint` and run `./start_sprint.sh` - (Windows): run `start_sprint.cmd` in `%userprofile%\Sites\sprint`
+4. Create Drupal 8 instance to use during sprint: (macOS/Linux): cd to `~/Sites/sprint` and run `./start_sprint.sh` - (Windows): run `start_sprint.cmd` in `%userprofile%\Sites\sprint`.
 5. Follow prompts from output of step 4 to start your Drupal 8 site.
-6. _Windows users must run an additional command_: the start_clean command will tell you what the command is.
+6. _Windows users_: To complete setup, run the `start_clean.cmd` command inside `%userprofile%\Sites\sprint`, and wait for it to complete (this will take a while). After the command completes:
```
mradcliffe: Is the start_clean.cmd is only mentioned here because start_clean.sh is displayed in the output for start_sprint? I think all users need to run start_clean as the next step. Is it necessary to also mention to run the equivalent for MacOS or Linux users?

mradcliffe (commented):
Should start_clean.sh be mentioned for Linux or MacOS users? Is that necessary since start_sprint.sh prints the output?

SPRINTUSER_README.md
BrianGilbert: Yes, everyone needs to run start_clean at least once regardless of OS.

BrianGilbert: @geerlingguy ``` start_clean.cmd` command inside `%userprofile%\Sites\sprint` ```

start_clean.cmd is inside the directory generated when you run start_sprint.cmd, not in the root sprint dir.

BrianGilbert (changes requested):
Step 6 needs rewording I think

geerlingguy:
@BrianGilbert - Changes made; I also slightly reorganized the layout of the instructions so it's more apparent which set of directions you should use for each step (it was a little annoying reading it all inline, so I added some structure).

Let me know what you think!

BrianGilbert (approved):

BrianGilbert:
@geerlingguy nice, that does seem clearer.. thanks.

PR #57, opened 2018-04-11
Build and test improvements

rfay:
* Change .latest_version to .ddev_version, add .quicksprint_version.txt
* Add a separate tarball which has no docker installs
* Add SHA256 hashes for the various tarballs.
* Move tests into a test script.


rfay:
This is probably OK as it is, please review. My intent is to add more tests to the test script.

BrianGilbert (approved):

Issue #61, opened 2018-04-13
Remove IRC container


BrianGilbert:
I was originally directed to include this but given only 1 person at the First Time Sprinters workshop is known to have used it, it would be much more economical for them to use irccloud or https://webchat.freenode.net/

mradcliffe:
Agreed. https://drupal.org/chat shows new contributors how to connect to the various platforms.

rfay:
This was removed in https://github.com/drud/quicksprint/pull/62

PR #62, opened 2018-04-13
[#61] Removed IRC container.

BrianGilbert:
Overwhelming majority of first time sprinters didn't use IRC, irrespective of if some people think it's still required much more economical to use IRCCloud or https://webchat.freenode.com


mradcliffe (commented):
I think there is still a reference in COPYING.

BrianGilbert:
@mradcliffe not anymore.. I'd already taken it out locvally.. not sure how I'd missed commiting it.

PR #63, opened 2018-07-21
Adds packaging instructions for sprint maintainers

mradcliffe:
I had this documentation written for several months now locally, but hadn't committed it.

Should also see if build is still passing in CI land.

### DIFF
--- a/MAINTAINER_README.md
+++ b/MAINTAINER_README.md
@@ -4,12 +4,50 @@ Quicksprint is a basic toolkit to get people started with ddev and a Drupal code
 
 There are two parts to this project:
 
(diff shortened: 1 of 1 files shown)


rfay (approved):
This is OK with me. This will eventually become the *DRUPAL_MAINTAINER_README.md* as this repo will be used for other projects as well. So I was a little concerned at first to see Drupal-specific stuff but then realized that this is still the original Drupal-oriented doc, so fine to go with this for now.

Thanks for this, and please let me know when you're ready to have it pulled. 

BTW, this repo has a nightly build every night, so that's good. It did break a couple of times showing missing items in ddev releases!

mradcliffe:
I think it's ready if approved.

I'll start ramping up testing the packages to see if there are any gaps for upcoming sprints.

+1 on name changes when we do the refactor to support TYPO3 and such.

rfay:
Awesome, thanks so much! This does need some love, so please let us know about upcoming sprints that may use it.

Issue #66, opened 2018-07-21
Update config.yaml to correct version of bundled ddev using "ddev config"


mradcliffe:
Update config.yaml for ddev v1.0.0:

```
Your /home/username/Sites/sprint/sprint-20180721-1509/.ddev/config.yaml version is 1, but ddev is version v1.0.0. 
Please run 'ddev config' to update your config.yaml. 
ddev may not operate correctly until you do. 
```

rfay:
It would probably be more sustainable just to run `ddev config` with args to update the config.yaml.

PR #67, opened 2018-07-22
Runs ddev config before ddev start and updates config.yaml (#66)

mradcliffe:
This addresses #66.

Tested shell script version on MacOS Sierra (works) and Linux (nginx didn't start without explicitly running as root):

- Without an explicit sudo (the shell script runs a sudo command), on MacOS, it is still treated as running as root since I still get the "Warning: containers will run as root. This could be a security risk on Linux", but I do not get that message on Slackware and the web container times out because nginx fails to start.
- With an explicit sudo, nginx does start, and docker does output the root warning above.

Todo:

- [x] Test on MacOS
- [x] Test on Linux (do a test on a more heavily used distro than the one I use)
- [x] Test on a Windows 10 desktop I have access to

### DIFF
--- a/sprint/.ddev/config.yaml
+++ b/sprint/.ddev/config.yaml
@@ -5,7 +5,7 @@ docroot: drupal8
 php_version: "7.2"
 router_http_port: "8080"
 router_https_port: "8443"
-xdebug_enabled: true
+xdebug_enabled: false
 additional_hostnames: []
 additional_fqdns: []
 provider: default
--- a/sprint/.ddev/config.yaml
+++ b/sprint/.ddev/config.yaml
@@ -1,17 +1,67 @@
-APIVersion: "1"
+APIVersion: v1.0.0
 name: sprint-[ts]
 type: drupal8
 docroot: drupal8
 php_version: "7.2"
 router_http_port: "8080"
 router_https_port: "8443"
+xdebug_enabled: true
+additional_hostnames: []
+additional_fqdns: []
 provider: default
 
-# Certain ddev commands can be extended to run tasks before or after the ddev
-# command is executed.
+# This config.yaml was created with ddev version v1.0.0
+# webimage: drud/ddev-webserver:v1.0.0
+# dbimage: drud/ddev-dbserver:v1.0.0
+# dbaimage: drud/phpmyadmin:v1.0.0
+# However we do not recommend explicitly wiring these images into the
+# config.yaml as they may break future versions of ddev.
+# You can update this config.yaml using 'ddev config'.
+
+# Key features of ddev's config.yaml:
+
+# name: <projectname> # Name of the project, automatically provides
+#   http://projectname.ddev.local and https://projectname.ddev.local
+
+# type: <projecttype>  # drupal6/7/8, backdrop, typo3, wordpress, php
+
+# docroot: <relative_path> # Relative path to the directory containing index.php.
+
+# php_version: "7.1"  # PHP version to use, "5.6", "7.0", "7.1", "7.2"
+
+# You can explicitly specify the webimage, dbimage, dbaimage lines but this
+# is not recommended, as the images are often closely tied to ddev's' behavior,
+# so this can break upgrades.
+
+# webimage: <docker_image>  # nginx/php docker image.
+# dbimage: <docker_image>  # mariadb docker image.
+# dbaimage: <docker_image>
+
+# router_http_port: <port>  # Port to be used for http (defaults to port 80)
+# router_https_port: <port> # Port for https (defaults to 443)
+
+# xdebug_enabled: false  # Set to true to enable xdebug and "ddev start" or "ddev restart"
+
+#additional_hostnames:
+# - somename
+# - someothername
+# would provide http and https URLs for "somename.ddev.local"
+# and "someothername.ddev.local".
+
+#additional_fqdns:
+# - example.com
+# - sub1.example.com
+# would provide http and https URLs for "example.com" and "sub1.example.com"
+# Please take care with this because it can cause great confusion.
+
+# provider: default # Currently either "default" or "pantheon"
+#
+# Many ddev commands can be extended to run tasks after the ddev command is
+# executed.
 # See https://ddev.readthedocs.io/en/latest/users/extending-commands/ for more
 # information on the commands that can be extended and the tasks you can define
-# for them.
-# hooks:
-#   post-import-db:
-#     - exec: "drush cr"
+# for them. Example:
+#hooks:
+# post-import-db:
+#   - exec: drush cr
+#   - exec: drush updb
--- a/sprint/start_clean.cmd
+++ b/sprint/start_clean.cmd
@@ -23,7 +23,7 @@ ECHO ####
 PAUSE
 
 REM Attempts to reconfigure ddev to update config automagically.
-ddev config default --docroot drupal8 --projecttype drupal8 --projectname sprint-[ts]
+ddev config --docroot drupal8 --projecttype drupal8 --projectname sprint-[ts]
 
 ddev start
 ddev exec git fetch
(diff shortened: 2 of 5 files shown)


sprint/.ddev/config.yaml
```diff
@@ -1,17 +1,67 @@
-APIVersion: "1"
+APIVersion: v1.0.0
 name: sprint-[ts]
 type: drupal8
 docroot: drupal8
 php_version: "7.2"
 router_http_port: "8080"
 router_https_port: "8443"
+xdebug_enabled: true
```
mradcliffe: Sorry, I added this as well.

mradcliffe:
Double-checked my dockerd options in my service to confirm I had a [docker group](https://github.com/drud/ddev/issues/750#issuecomment-376500277) (didn't in my latest docker build, which I had previously as my local user was in the docker group I created), restarted docker, but that didn't have an effect.

sprint/.ddev/config.yaml
rfay: I don't think this should be there by default, slows down performance quite a lot.

rfay:
There is no place anybody should be using sudo with ddev. 

Linux users who haven't used docker before often stumble by not doing the [post config steps](https://docs.docker.com/install/linux/linux-postinstall/#manage-docker-as-a-non-root-user), which are called out repeatedly in the [ddev docs - system requirements](https://ddev.readthedocs.io/en/latest/), and are fundamental. Mostly putting the local user in the docker group. 

But it trips up loads of people. In the next release we plan to improve the linux install docs, https://github.com/drud/ddev/issues/1002. But I kind of think ddev should check on linux, or perhaps in the install script? to see if this has been done.

mradcliffe:
I don't have any issues with running docker as my own user currently (without sudo) on other containers so I think there is something else specific to nginx or the ddev-nginx-php container that is causing this on my system. But more importantly, testing the PR on a more standard Linux distro. is probably a better idea.

Found some issues on the current Windows build.

- start_sprint.cmd references `bin\7za.exe` and `bin\sed.exe` when it is now in the `bin\windows\` directory.
- start_clean.cmd doesn't work for Docker Toolbox (wasn't designed for it because ddev didn't support it back in April) so the check for docker running doesn't work. This should be a follow-up IMO.
- git operations are failing inside Docker Toolbox because the git URL change went through last night and we're using git://git.drupal.org. :|

Otherwise quicksprint and ddev are working great on Windows 10 Pro running Docker Toolbox! 👍

sprint/.ddev/config.yaml
mradcliffe: I'll revert it. I changed it because it would be nice as default during a sprint, but then again, not everyone is going to contribute or debug code at a sprint.

mradcliffe:
I noticed that `test_drupal_quicksprint.sh` on circleci master/nightly as well as in the build in this PR on CircleCI is broken and not actually testing or failing the build. It may not have ever been working? I'm checking it out.

```
chmod: cannot access ‘sprint-*’: No such file or directory
./test_drupal_quicksprint.sh: line 22: cd: /home/circleci/Sites/sprint/sprint-2*: No such file or directory
Test successful
```

rfay:
Nice work, thanks. I think we're going to schedule some time for Quicksprint in our next sprint, too.

test_drupal_quicksprint.sh
```diff
@@ -15,12 +15,15 @@ tar -C /tmp -zxf ~/tmp/drupal_sprint_package.no_docker.$(cat ./.quicksprint_rele
 # Run the standard install_ddev.sh, asserting we already have docker set up.
 cd /tmp/drupal_sprint_package && echo y | ./install_ddev.sh
 
+
 # Remove any existing sprints, then do start_sprint.sh
-cd ~/Sites/sprint && chmod -R u+w sprint-* && rm -rf sprint-* echo y | ./start_sprint.sh
+cd ~/Sites/sprint
+find ~/Sites/sprint -type d -name "sprint-2*" -exec chmod -R u+w '{}' \; -exec rm -rf '{}' \;
```
mradcliffe, line 25: I always have trouble using `xargs`, which is more efficient than `-exec`.

test_drupal_quicksprint.sh
```diff
@@ -15,12 +15,15 @@ tar -C /tmp -zxf ~/tmp/drupal_sprint_package.no_docker.$(cat ./.quicksprint_rele
...
+echo y | ./start_sprint.sh
 
 # Run start_sprint in the (only) sprint directory
 cd ~/Sites/sprint/sprint-2* && echo y | ./start_clean.sh
```
mradcliffe, line 29: There's probably some exit 0/1 assertion that can be done afterward.

Maybe check ddev health again?

sprint/start_clean.cmd
```diff
@@ -22,6 +22,9 @@ ECHO #
 ECHO ####
 PAUSE
 
+REM Attempts to reconfigure ddev to update config automagically.
+ddev config default --docroot drupal8 --projecttype drupal8 --projectname sprint-[ts]
```
rfay: "default" isn't necessary here.

sprint/start_clean.cmd
```diff
@@ -22,6 +22,9 @@ ECHO #
 ECHO ####
```
rfay, line 22: maintaining cmd scripts is quite a burden, and it's a horrible scripting language to boot. My strategy I think is to install/require git-bash for all windows users, and use bash scripts that are identical for the different operating systems.

sprint/start_clean.sh
```diff
@@ -43,6 +43,10 @@ while true; do
     esac
 done
 
+
+# Attempts to reconfigure ddev to update config automagically.
+ddev config default --docroot drupal8 --projectname sprint-[ts] --projecttype drupal8
```
rfay: Again, default not required.

test_drupal_quicksprint.sh
rfay: xargs is nice, `find ~/Sites/sprint -type d -maxdepth 1 -name "sprint-2*" | xargs rm -r` ; that doesn't solve the fact that you want to do both the chmod and the rm.

test_drupal_quicksprint.sh
```diff
@@ -15,12 +15,23 @@ tar -C /tmp -zxf ~/tmp/drupal_sprint_package.no_docker.$(cat ./.quicksprint_rele
...
 
+# Confirms web and router status.
+ddev describe | grep running || { echo "ERROR: ddev web is not running correctly?"; exit 1; }
+ddev describe | grep "DDEV ROUTER STATUS: healthy" || { echo "ERROR: ddev router status is not healthy?" ; exit 1; }
```
rfay: Just for grins, it's probably better to use a different exit status for each item.

Also, if we're using `set -o errexit pipefail` none of this is necessary (but this is far more explicit and excellent)

rfay (commented):
Thanks for your work on this; minor comments and suggestions. Looks pretty close to ready. I'm expecting to spend some time on this in the next couple of weeks as well.

sprint/start_clean.cmd
mradcliffe: I think initially Brian was looking babun (and docksal). Babun would be nice, but increases package size by a bunch. Git Bash is a bit limited, but decent enough though requires more extra steps for Windows users. Powershell would be nice too, despite the additional burden, but not bundled with Windows 7 by default (not sure about 10 Home).

That written, :+1: to consolidation.

test_drupal_quicksprint.sh
mradcliffe: I took a stab at some "constants" for exit statuses, but I think it's wrong. I should have used either less than 128 or less than 255. Whoops. Fix for that soon.

rfay:
Where are you at with this PR? Can you rebase it please?  Thanks!

rfay:
I used github's "update from master" feature to get this updated. Now that I've used it once I don't approve of it :) But OTOH you don't have to rebase!

rfay:
Thanks for your work on this. I know it's not all the way yet, but moving along.

PR #69, opened 2018-07-23
Changes to using git HTTPS URL per https://drupal.org/gitauth (#69)

mradcliffe:
https://drupal.org/gitauth

### DIFF
--- a/package_drupal_script.sh
+++ b/package_drupal_script.sh
@@ -148,7 +148,7 @@ done
 
 # clone or refresh d8 clone
 mkdir -p sprint
-git clone git://git.drupal.org/project/drupal.git $STAGING_DIR/sprint/drupal8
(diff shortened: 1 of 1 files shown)


rfay:
I don't see anything in that article about git:// being changed.  It remains one of the best unauthenticated and reliable techniques for public repos.

mradcliffe:
Hmm, it's the firewall where I am at blocking git URLs.

That also could be a good reason to change to either `ssh://git@git.drupal.org` (new pattern) or `https://git.drupal.org`.

rfay:
ssh is problematic, won't work for lots of people, and is blocked by lots of firewalls.
You win on https, because it is overall better for firewalls.  So 👍 for https urls.

rfay (approved):
This is fine then, switching to https should make it work in a majority of cases. I was going to say it didn't really matter because this is *packaging* time, but the *remote* then gets used by our users.

PR #71, opened 2018-08-15
Remove unused cloud 9 additions

rfay:
Cloud 9 ended up not being useful, and it wasn't clear that anybody used it. To simplify the package and reduce the docker resource load, remove it.


mradcliffe:
I think that we (mentors) still would like to distribute some sort of editor. Cloud 9 seemed like a good idea, but I think the limitations with not being able to figure out XDebug probably aren't worth the image size.

I'm a VS Code fan (when I'm on MacOS or Windows without my personal PHPStorm license), and I think that might be the way to go though distribution through the MIT License might only be possible by building the source directly. That would mean packaging each of the OS versions (maybe not for Linux the same as docker distribution currently). There might be some other folk doing this already.

rfay:
I don't really have an objection to adding a VSCode distribution or something. Might even be OK to give people PHPStorm (30-day eval). 

Off-topic... your PHPStorm license is good as many places as you need it. I have at least 10 activations right now :)

andrewfrench (approved):
The removals here look reasonable to me for removing Cloud9.

Issue #74, opened 2018-08-15
Rewrite/improve testing (test_drupal_quicksprint.sh doesn't actually work)


rfay:
As @mradcliffe pointed out, test_drupal_quicksprint.sh isn't actually working, and it's unclear how the build is passing despite that. See https://circleci.com/gh/drud/quicksprint/180 - the test clearly fails.

I'd like to redo this perhaps with [BATS](https://github.com/sstephenson/bats)

rfay:
Fixed in #75

Issue #73, opened 2018-08-15
docker 18.03 versions are wired into scripts


rfay:
18.03 is explicitly mentioned a number of places;  but v1.1.0 requires 18.06, and we probably want to figure out how to globally specify version or *something*.

rfay:
Fixed in #75

Issue #68, opened 2018-07-22
Add macOS and Windows builds/tests


rfay:
This project would be much more robust with macOS and Windows builds/tests. We currently build on Linux, which is great, but if we could figure out how to run everything on the other platforms it would be a step forward. We're using buildkite for ddev tests on macOS, Win10Pro, Win10Home, we could do the same here.

rfay:
Fixed in #75 - we now have testing on macOS and docker toolbox and docker-for-windows.

Issue #64, opened 2018-07-21
Bundle Windows installer, Update instructions to use the installer


rfay:
Since ddev now has a Windows installer, there are a number of things that should be updated:

- [x] Bundle the Windows installer
- [x] Update docs to tell people to run the installer (instead of the cmd script)
- [x] Remove cmd installer scripts

rfay:
The installer is now bundled, but I want to take a from-scratch go-through on this. I think this is the most important outstanding item (or experience) remaining before we do a release.

rfay:
This was successful, so closing.

PR #75, opened 2018-08-15
Rework tests as bats tests

rfay:
This reworks the tests from a naive bash script to one that uses [bats](https://github.com/bats-core/bats-core). Hopefully it will be more effective. It aims at the same basic functionality as the original test script was doing.

I apologize that this got a bit out of hand. I wanted to make the tests work on WIndows and Mac as well as Linux. And one thing led to another. There were lots of changes required in the scripts for that to work, especially for windows. And I ended up throwing away the *.cmd scripts in this PR, which was intended to be later.


tests/installation.bats
```diff
@@ -0,0 +1,73 @@
...
+    # Make sure passwordless sudo is available
+    echo junk | sudo -S ls
+    docker run -t -v "$HOME:/tmp/home" -p 80:8088 busybox ls >/dev/null
+}
+
+
+@test "untar and run drupal_sprint_package" {
+    rm -rf "$UNTAR_LOCATION/drupal_sprint_package"
+    tar -C "$UNTAR_LOCATION" -zxf "$SOURCE_TARBALL_LOCATION"
+    chmod -R ugo+w $SPRINTDIR/ && rm -rf $SPRINTDIR/sprint-2*
```
mradcliffe: Probably doesn't exist before install_ddev.sh. I left this logic alone in the test because I'm not familiar with circle ci, but probably circle ci runs everything in a fresh environment so maybe this isn't necessary?

rfay: Yeah, thanks. I'm definitely going to make that not fail, as it's optional.

SPRINTUSER_README.md
```diff
@@ -19,16 +19,10 @@ Installation and Startup:
 2. Start Docker (it must be running to install ddev)
 3. Open Docker preferences and set docker memory allocation to 3.0GB in the Advanced section (required for container to be started)
 3. Install DDEV-Local:
-  - **macOS/Linux**: Run install_ddev.sh via terminal from the directory where this SPRINTUSER_README.md is with `./install_ddev.sh`
-  - **Windows**: Run `install_ddev.cmd`
+  - Run install_ddev.sh via terminal (git-bash on Windows)from the directory where this SPRINTUSER_README.md is with `./install_ddev.sh`
```
mradcliffe, line 22: Maybe we could remove "(git-bash on Windows)" from this bullet, and then add a sub-bullet for Windows users?

```
   - Windows 10 Professional users should run "Git Bash".
   - Windows 7, Windows 10 Home users should run "Docker Quickstart Terminal".
```

install_ddev.sh
```diff
@@ -46,46 +46,50 @@ ${GREEN}
...
-fi
-
 echo ""
 echo "Installing docker images for ddev to use..."
-if [[ "$OS" == "Darwin" ]]; then
-    gzip -dc $(ls ddev_tarballs/ddev_docker_images*.tar.xz) | docker load
-elif [[ "$OS" == "Linux" ]]; then
-    xzcat $(ls ddev_tarballs/ddev_docker_images*.tar.xz) | docker load
-fi
+7z x ddev_tarballs/ddev_docker_images.*.tar.xz -so | docker load
```
mradcliffe: I'm okay with removing the commands for Windows users and relying on Git Bash or Docker Quickstart Terminal.

A year ago I ran into issues with trying to get 7zip to run in those environments on my own docker-core-sprint, but it's possible that Git Bash (with or without Docker Quickstart Terminal) did not include it at that point (?). Need to test this on Docker Toolbox.

mradcliffe: I think that Powershell is probably preferable for Window 10 Pro users, but I don't think we can support that because of the numerous Windows 10 versions and whatever they support.

install_ddev.sh
```diff
@@ -46,46 +46,50 @@ ${GREEN}
...
+TARBALL=""
+case "$OS" in
+    Linux)
+        TARBALL=ddev_tarballs/ddev_linux*.tar.gz
+        ;;
+    Darwin)
+        TARBALL=ddev_tarballs/ddev_macos*.tar.gz
+        ;;
+    MINGW64_NT-10.0)
+        echo "${YELLOW}PLease use the ddev_windows_installer provided with this package to install ddev${RESET}"
```
mradcliffe: The prompt on line 39 probably needs to be updated to mention that Windows users should install ddev. At the moment it mentions that the script installs ddev for them.

package_drupal_script.sh
```diff
@@ -17,16 +17,13 @@ export GZIP=-9
```
mradcliffe, line 17: The changes here make sense:

- Improve bash variable use with quoted strings.
- Using pushd/popd.
- Removing unnecessary conditionals.

start_sprint.sh
```diff
@@ -15,22 +15,12 @@ OS=$(uname)
...
-	sed -i '' 's/\[ts\]/'${TIMESTAMP}'/' sprint-$TIMESTAMP/Readme.txt
-	sed -i '' 's/\[ts\]/'${TIMESTAMP}'/' sprint-$TIMESTAMP/start_clean.sh
-	sed -i '' 's/\[ts\]/'${TIMESTAMP}'/' sprint-$TIMESTAMP/start_clean.cmd
-elif [[ "$OS" == "Linux" ]]; then
-	sed -i 's/\[ts\]/'${TIMESTAMP}'/' sprint-$TIMESTAMP/.ddev/config.yaml
-	sed -i 's/\[ts\]/'${TIMESTAMP}'/' sprint-$TIMESTAMP/Readme.txt
-	sed -i 's/\[ts\]/'${TIMESTAMP}'/' sprint-$TIMESTAMP/start_clean.sh
-	sed -i 's/\[ts\]/'${TIMESTAMP}'/' sprint-$TIMESTAMP/start_clean.cmd
-fi
+perl -pi -e "s/\[ts\]/${TIMESTAMP}/g" sprint-${TIMESTAMP}/* sprint-${TIMESTAMP}/.ddev/config.yaml
```
mradcliffe: It looks like perl is included in git-bash 2.x versions. :+1:

install_ddev.sh
```diff
@@ -40,7 +37,7 @@ ${GREEN}
 # It does the following:
 #  -Install Docker for your OS if you don't have it already
 #  -Install ddev by Drud Technology
-#  -Copy required components to ~/Sites/sprint/
```
mradcliffe: Above here, we should note that this does not install ddev automatically for Windows users.

SPRINTUSER_README.md
mradcliffe: And that Windows users need to install ddev first, before running install_ddev.sh. That's a little confusing so maybe should rename the script "install.sh"?

tests/installation.bats
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env bats
```
mradcliffe, line 1: Nice. I've never heard about BATS before, and it looks really interesting. :+1:

mradcliffe (changes requested):
The bash script changes look fine to me, and :+1: on the better tests.

There are two things that are probably remaining here:
1. Give it a go on Windows. I'll run it through a Docker Toolbox and Docker Quickstart Terminal to ensure that I can run 7zip and perl through that instance of Git-Bash.
2. Update some of the documentation for Windows users regarding the changes to installing ddev.

I think that we should create a follow-up to rename install_ddev.sh to install.sh because this changeset is quite large.

SPRINTUSER_README.md
mradcliffe: The prerequisites section of SPRINTUSER_README is inaccurate (and has been since the latest ddev v1 release that supports older Windows versions).

rfay: Both great points, thanks. The docs will have to change quite a lot, so I'm not as worried about that, but I'm glad you called it out. git bash and DQT are in fact the same thing AFAICT, but you're right that we should be clear on that. In my experience, git-bash works great on win10home/dockertoolbox. 

Definitely have to rename the script.

install_ddev.sh
rfay: I'm not even sure we should automatically install it for anybody, thinking...

rfay: I'm pretty sure the fact this doesn't automatically install will mess up the windows test, so still thinking about that too.

tests/installation.bats
rfay: I'm doing a pretty bad use here, because the tests build on each other, which is not what you're supposed to do, they're supposed to be completely independent of each other. But the results are much easier to read than the results of a script that just terminates when something breaks.

I've debated writing a better test script for the main build and install and then a little bit of bats to check and make sure things are working. That would be more like the intent of bats, but this does work mostly ok.

install_ddev.sh
mradcliffe: On Docker Toolbox, Git Bash doesn't have a /usr/local/bin so BINOWNER fails. I guess we can't use `whoami` here because it's for moving the ddev binary into place. However, since this is Windows, and we're installing ddev via the installer anyway, this could be ignored/error caught.

```bash
USERNAME@MACHINE MINGW64 /c/Users/USERNAME/Desktop/drupal_sprint_package
$ ./install_ddev.sh
ls: cannot access '/usr/local/bin': No such file or directory
```

install_ddev.sh
```diff
@@ -49,106 +46,55 @@ ${GREEN}
...
-elif [[ "$OS" == "Linux" ]]; then
-    xzcat $(ls ddev_tarballs/ddev_docker_images*.tar.xz) | docker load
-fi
-
-if [ -f ddev_tarballs/docker_additions.tar.xz ]; then
-    if [[ "$OS" == "Darwin" ]]; then
-        gzip -dc $(ls ddev_tarballs/docker_additions.tar.xz) | docker load
-    elif [[ "$OS" == "Linux" ]]; then
-        xzcat $(ls ddev_tarballs/docker_additions.tar.xz) | docker load
+7z x ddev_tarballs/ddev_docker_images.*.tar.xz -so | docker load
```
mradcliffe: On Windows 10, Docker Toolbox, Docker Quickstart Terminal:
```bash
Continue? (y/n): y


Installing docker images for ddev to use...
./install_ddev.sh: line 58: 7z: command not found
```

But the docker toolbox version I got in the package was 18.03ce.

package_drupal_script.sh
```diff
@@ -13,16 +17,13 @@ export GZIP=-9
...
 QUICKSPRINT_RELEASE=$(git describe --tags --always --dirty)
 
-echo $QUICKSPRINT_RELEASE >.quicksprint_release.txt
+echo "$QUICKSPRINT_RELEASE" >.quicksprint_release.txt
 
-# The version lines on the following few lines need to get changed any time the url are changed on the line below.
-DOCKER_URLS="https://download.docker.com/mac/stable/23751/Docker.dmg https://download.docker.com/win/stable/16762/Docker%20for%20Windows%20Installer.exe"
-DOCKER_VERSION_MAC="18.03.0-ce-mac60"
-DOCKER_VERSION_WIN="18.03.0-ce-win59"
+DOCKER_URLS="https://download.docker.com/mac/stable/Docker.dmg https://download.docker.com/win/stable/Docker%20for%20Windows%20Installer.exe https://download.docker.com/win/stable/DockerToolbox.exe"
```
mradcliffe: Docker Toolbox installer from here was 18.03ce not 18.06ce.

install_ddev.sh
mradcliffe: 7zip doesn't appear to be included in any version of Git for Windows, but msys2 packages could be installed with it.

Adding PATH=$PATH:$PWD/bin/windows before the command arnd renaming bin/windows/7za.exe works for me.

install_ddev.sh
mradcliffe: The output of this line got messed up with the docker load output making it hard to read so maybe add a line break before it.

start_sprint.sh
```diff
@@ -15,22 +15,12 @@ OS=$(uname)
 TIMESTAMP=$(date +"%Y%m%d-%H%M")
 
 #Extract a new ddev D8 core instance to $CWD/sprint-$TIMESTAMP
-mkdir -p sprint-$TIMESTAMP
-tar xpf sprint.tar.xz -C sprint-$TIMESTAMP
+mkdir -p sprint-${TIMESTAMP}
+tar xpf sprint.tar.xz -C sprint-${TIMESTAMP}
```
mradcliffe: Reminder to create a follow-up issue to echo some output before extracting the sprint tarball to ask for patience. Also could help for git operations in start_clean.sh.

start_sprint.sh
mradcliffe:
```
$ ./start_sprint.sh
Can't do inplace edit: sprint-20180818-1505/drupal8 is not a regular file.
```

Also perl probably is generating some backup files:
```
$ ls
drupal8/  Readme.txt  Readme.txt.bak  start_clean.sh*  start_clean.sh.bak*
```

Both of these did not have any effect on the outcome.

sprint/start_clean.sh
```diff
@@ -4,19 +4,15 @@ set -o errexit
...
 YELLOW='\033[33m'
 RESET='\033[0m'
 
 # Check Docker is running
-SERVICE='docker'
-if ps ax | grep -v grep | grep -v /Library/PrivilegedHelperTools/com.docker.vmnetd | grep $SERVICE > /dev/null
+if docker run --rm -t busybox:latest ls >/dev/null
 then
-    printf "${GREEN}$SERVICE service running, continuing.\n${RESET}"
+    printf "docker service running, continuing."
```
mradcliffe, line 15: Not related to this line, but we forgot to remove the IDE line in the comments for #71.

```
 # ${GREEN}IDE:       ${YELLOW}http://sprint-[ts].ddev.local:8000/${GREEN}
 #            ${YELLOW}(U:username  P:password)${GREEN}
 #
```

mradcliffe (changes requested):
Summary:

1. BINOWNER probably needs to be moved to the the Linux/MacOS sections as it fails on Git-Bash.
2. Add bin/windows PATH to 7z command and rename 7za.exe to 7z.
3. Line break before "Please use the ddev_windows_installer..." text for output readability.
4. Update Docker Toolbox URL because Docker's simple URL For docker toolbox isn't up-to-date.

install_ddev.sh
rfay: I added 7z to the path and renamed it; also tried some other magic to get the most possibilities we could. Not sure the use of xz was a good idea. Lots of hoop-jumping here.

rfay:
I think I got all your comments taken care of, thanks @mradcliffe - And we're close to green on mac, linux, and both windows. The one thing that is a little bit of a cheat is that the testbots have 7zip and zip installed, for example.

sprint/start_clean.sh
rfay: Removed, thanks.

rfay:
Yay GREEN! 💚

rfay:
I'm going to go ahead and pull because I think all @mradcliffe's requests are in, and because we're in a time of rapid change so if he has any regrets about what we have in this one we can get them in still.

mradcliffe:
Ok, this looks good.

mradcliffe:
This PR addresses #73, #74, #68, #64, and #60 .

PR #78, opened 2018-08-18
Update docs for windows-git and bundle windows git installer, fixes #65

### LINKED ISSUE #65: Update support info about Windows 10 Home
Since ddev now supports Windows 10 Home with Docker Toolbox, the docs here need to point to how to do that.

rfay:
* Update user docs in general
* Specific updates for Windows, explaining new ddev windows installer, etc.
* Bundle git for Windows in the package.


rfay:
I think this one is good, and we seem to have green on Windows. I think @mradcliffe would propose that we do a release after this, then proceed to some simplification of the scripts. (Get it down to just a packaging script (for maintainers), an install script, and a start script, although the start script should probably just be "ddev start", so we'll want to use the script to prompt them probably.

mradcliffe:
Shoot, I have some docs improvements that are going to conflict here, @rfay . Was about to open a PR, but I can rebase after this.

mradcliffe (approved):
Looks good to me.

PR #80, opened 2018-08-19
Improves documentation for sprint users

mradcliffe:
This is pretty similar to what I used for docker-core-sprint, and I had some mentors run through and give suggestions for Nashville.

Now rebased directly off of upstream/master.

### DIFF
--- a/SPRINTUSER_README.md
+++ b/SPRINTUSER_README.md
@@ -10,8 +10,8 @@ This directory contains tools to get you started contributing to Drupal 8:
 ## Prerequisites
 
 * A computer with 8gb memory or greater.
-* Windows 7 or higher, MacOS Sierra or higher or a recent/stable Linux distribution.
-* A robust code editor such as Visual Studio Code, Atom, PhpStorm, Netbeans, vim or emacs (this may be provided as part of this package).
+* Windows 7 or higher, MacOS El Capitan or higher or a recent/stable Linux distribution.
+* A robust code editor such as Visual Studio Code, Atom, PhpStorm or Netbeans (this may be provided as part of this package).
 
 **⚠️ You can still install and contribute to Drupal even without these requirements using [Drupal's quick-start command](https://www.drupal.org/docs/8/install/quick-start-launch-a-local-demo-version-of-drupal-8-using-4-brief-steps)**!
 
@@ -33,8 +33,7 @@ This directory contains tools to get you started contributing to Drupal 8:
 <a name="extract-files"></a>
 ### 2. Extract drupal_sprint_package directory
 
-1. Extract the drupal_sprint_package-RELEASE.zip file to your Desktop. It is important that these files exist within your "user directory".
-2. Browse to this directory.
+Extract the drupal_sprint_package-RELEASE.zip file, and open or browse to the "drupal_sprint_package" directory.
 
 <a name="install"></a>
 ### 3. Install Docker and Other Requirements
--- a/SPRINTUSER_README.md
+++ b/SPRINTUSER_README.md
@@ -1,32 +1,94 @@
-# Drupal 8 Sprint Package
+# Drupal 8 Contribution Package
 
-This directory contains tools to get you going at a Drupal 8 sprint:
+This directory contains tools to get you started contributing to Drupal 8:
 
-* Drupal 8, already cloned with git (`git clone git://git.drupal.org/project/drupal.git`)
-* Docker (you must install it yourself; install images are provided in the installs directory). Instructions and troubleshooting information at https://ddev.readthedocs.io/en/latest/users/docker_installation/
-* Git For Windows (for Windows Users)
+* Drupal 8, already cloned with git
+* Docker CE
 * DDEV-Local (ddev) development environment
+* Additional tools for Windows
 
-Prerequisites:
+## Prerequisites
 
-**Windows Prerequisites:**
-* Windows 10 Professional/Enterprise: Please use Docker for Windows (bundled); Hyper-V must be enabled. If you don't have these, the Docker install process will typically prompt for the settings. 
-* Windows 10 Home (and other versions like Server): Please use Docker Toolbox (bundled). You must start up Docker Toolbox using the "Docker Toolbox Quickstart"
-* [Git for Windows](https://gitforwindows.org/): git-bash is used for all installation scripts, and of course you'll need git to be able to develop patches. [Helpful git setup info at github](https://help.github.com/articles/set-up-git/). 
-* ddev Windows Installer (bundled)
+* A computer with 8gb memory or greater.
+* Windows 7 or higher, MacOS Sierra or higher or a recent/stable Linux distribution.
+* A robust code editor such as Visual Studio Code, Atom, PhpStorm, Netbeans, vim or emacs (this may be provided as part of this package).
 
-**macOS and Linux Prerequisites**
-* Docker-for-Mac (bundled) or for Linux. See [detailed instructions](https://ddev.readthedocs.io/en/latest/users/docker_installation/)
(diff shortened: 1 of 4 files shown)


SPRINTUSER_README.md
```diff
@@ -1,32 +1,94 @@
...
-3. Install DDEV-Local:
-  - Run install_ddev.sh via terminal (git-bash on Windows) from the directory where this SPRINTUSER_README.md is with `./install_ddev.sh` - On Windows also run the ddev_windows_installer bundled.
-4. Create Drupal 8 instance to use during sprint:
-  - cd to `~/sprint` and run `./start_sprint.sh`
-5. There should be a new sprint folder named `sprint-YYYYMMDD-HHMM` (with today's date and the time you ran `start_sprint`). Start the Drupal environment by using cd to move to the dated sprint folder, and run `./start_clean.sh`
-6. After start_clean is finished, it will provide you with a set of URLs and further instructions for using your sprint environment.
+## Prerequisites
+
+* A computer with 8gb memory or greater.
+* Windows 7 or higher, MacOS Sierra or higher or a recent/stable Linux distribution.
```
rfay: We haven't ever tried Windows 7, but it's likely to work. On macOS seems to be El Capitan, https://docs.docker.com/docker-for-mac/install/#what-to-know-before-you-install

SPRINTUSER_README.md
```diff
@@ -1,32 +1,94 @@
...
+
+ Operating System | Docker Version | Installer
+ ---------------- | -------------- | -----------------
+ Windows 10 Pro, Enterprise (HyperV enabled) | Docker CE | "Docker for Windows Installer.exe"
+ Windows 10 Home | Docker Toolbox | DockerToolbox-VERSION-ce.exe
+ Windows 7 (or no HyperV enabled)| Docker Toolbox | DockerToolbox-VERSION-ce.exe
+ MacOS | Docker CE | Docker.dmg
+ Linux | Docker CE, docker-compose | See [Linux instructions](https://docs.docker.com/install/#docker-ce)
+
+**⚠️ Docker Toolbox users**: Please install "Git for Windows" and "VirtualBox" as needed during the install process.
```
mradcliffe, line 53: I'm not sure, now that 2.18.0 is included, whether we should explicitly tell users to uncheck Git for Windows as I'm not sure if that will install Quickstart terminal. Reinstalling is fine. I tested that today.

SPRINTUSER_README.md
```diff
@@ -1,32 +1,94 @@
...
-  - Run install_ddev.sh via terminal (git-bash on Windows) from the directory where this SPRINTUSER_README.md is with `./install_ddev.sh` - On Windows also run the ddev_windows_installer bundled.
-4. Create Drupal 8 instance to use during sprint:
-  - cd to `~/sprint` and run `./start_sprint.sh`
-5. There should be a new sprint folder named `sprint-YYYYMMDD-HHMM` (with today's date and the time you ran `start_sprint`). Start the Drupal environment by using cd to move to the dated sprint folder, and run `./start_clean.sh`
-6. After start_clean is finished, it will provide you with a set of URLs and further instructions for using your sprint environment.
+## Prerequisites
+
+* A computer with 8gb memory or greater.
+* Windows 7 or higher, MacOS Sierra or higher or a recent/stable Linux distribution.
+* A robust code editor such as Visual Studio Code, Atom, PhpStorm, Netbeans, vim or emacs (this may be provided as part of this package).
```
rfay: Now I remember we were thinking about doing that... I don't think it's worth mentioning Netbeans any more. And not sure that anybody who needs this package would know how to use vim or emacs, but could be wrong.

SPRINTUSER_README.md
```diff
@@ -1,32 +1,94 @@
...
+### 1. Get The Files
+
+* USB Drive - Copy drupal_sprint_package-RELEASE.zip and Docker installer for your Operating System from the USB drive to your Desktop.
+* ResilioSync - Find the folder/directory that was downloaded and copy its contents to your Desktop.
+* GitHub - Download drupal_sprint_package-RELEASE.zip from https://github.com/drud/quicksprint/releases.
+
+<a name="extract-files"></a>
+### 2. Extract drupal_sprint_package directory
+
+1. Extract the drupal_sprint_package-RELEASE.zip file to your Desktop. It is important that these files exist within your "user directory".
```
rfay: Why is that important? I usually do it in /tmp...

SPRINTUSER_README.md
```diff
@@ -1,32 +1,94 @@
...
+
+#### 3.1 Docker CE or Docker Toolbox
+
+Find the Docker installer for your Operating System underneath the **installs** directory. It is important to install the version of Docker provided for compatibility with the tools.
+
+ Operating System | Docker Version | Installer
+ ---------------- | -------------- | -----------------
+ Windows 10 Pro, Enterprise (HyperV enabled) | Docker CE | "Docker for Windows Installer.exe"
+ Windows 10 Home | Docker Toolbox | DockerToolbox-VERSION-ce.exe
+ Windows 7 (or no HyperV enabled)| Docker Toolbox | DockerToolbox-VERSION-ce.exe
```
rfay, line 49: Docker Toolbox is *not* called "-ce"

sprint/start_clean.sh
```diff
@@ -64,6 +64,8 @@ ${GREEN}
 #
 # DB Admin:  ${YELLOW}http://sprint-[ts].ddev.local:8036/${GREEN}
```
rfay: "DB Admin" sounds awfully stiff and formal, somewhere else I changed this to "phpMyAdmin:"

rfay (approved):
I like this, requested some changes but you can respond as you like. I suspect would should pull the Windows 7 references because we don't have any way to support them.

SPRINTUSER_README.md
rfay: Oh, good point. Do this as it works. I forgot that it automatically includes that.

SPRINTUSER_README.md
mradcliffe: Ah, nice. I changed to El Capitan.

SPRINTUSER_README.md
mradcliffe: I removed vim/emacs, but I kept Netbeans as it is also a viable OS choice and still works well with XDebug and ES6. Although the intellisense is not as good as PhpStorm it allows to open multiple projects in the same window without mucking about and joining the projects together like PhpStorm. I use them both. It's only an option.

SPRINTUSER_README.md
mradcliffe: Whoops, I carried this over from old documentation in docker-core-sprint where I was running docker directly from the package directory (so it needed to be in the user directory for Windows and MacOS). I re-worded.

SPRINTUSER_README.md
mradcliffe: Agreed, it's not really CE, but the installer is still named as `VERSION-ce.exe`: `https://github.com/docker/toolbox/releases/download/v18.06.0-ce/DockerToolbox-18.06.0-ce.exe`

sprint/start_clean.sh
mradcliffe: Updated. I needed to change the spaces in all the lines by an additional space.

mradcliffe:
I'm satisfied if you are.

SPRINTUSER_README.md
rfay: You are right! Sorry, I thought they were different codepaths.

SPRINTUSER_README.md
rfay: That's fine. Last time I tried to use netbeans I failed so removed from ddev docs. Seemed like it wasn't getting any love at all.

rfay:
I'll pull this whenever you say it's ready.

rfay:
Just waiting for required circle build and will get it. I know that nothing in here can break it :)

Issue #77, opened 2018-08-18
Add echo statements to potential long-running commands


mradcliffe:
Follow-up from #75,

* start_sprint.sh
   * `tar xpf sprint.tar.xz`
* start_clean.sh (or the equivalent commands in start_sprint.sh if it still exists)
   * `ddev exec git fetch`

mradcliffe:
Both things are fixed now.

PR #81, opened 2018-08-19
Improve tests, bats tests should not include setup (or state)

rfay:
Up to now, the bats test were dependent on each other, but we need to make them just dependent on their own setup. This moves the basic build and install into the test script.

### DIFF
--- a/package_drupal_script.sh
+++ b/package_drupal_script.sh
@@ -149,7 +149,7 @@ cd ${STAGING_DIR}
 
 echo "Creating tar and zipballs"
 # Create tar.xz archive without using xz command, so we can work on all platforms
-pushd sprint >/dev/null && 7z a -ttar -so bogusfilename.tar . | 7z a -si -txz ../sprint.tar.xz >/dev/null && popd >/dev/null
+pushd sprint >/dev/null && 7z a -ttar -so bogusfilename.tar . 2>/dev/null | 7z a -si -txz ../sprint.tar.xz >/dev/null && popd >/dev/null
 rm -rf ${STAGING_DIR}/sprint
 
 cd ${STAGING_DIR_BASE}
--- a/package_drupal_script.sh
+++ b/package_drupal_script.sh
@@ -48,12 +48,7 @@ fi
 if [ -d "$STAGING_DIR" ] && [ ! -z "$(ls -A "$STAGING_DIR")" ] ; then
     printf "${RED}The staging directory $STAGING_DIR already has files. Deleting them and recreating everything.${RESET}"
     rm -rf "$STAGING_DIR"
-    if [ -e "$STAGING_DIR_BASE/drupal_sprint_package$QUICKSPRINT_RELEASE.tar.gz" ] ; then
-        rm "$STAGING_DIR_BASE/drupal_sprint_package$QUICKSPRINT_RELEASE.tar.gz"
-    fi
-    if [ -e "$STAGING_DIR_BASE/drupal_sprint_package$QUICKSPRINT_RELEASE.zip" ]; then
-        rm "$STAGING_DIR_BASE/drupal_sprint_package$QUICKSPRINT_RELEASE.zip"
-    fi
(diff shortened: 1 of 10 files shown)


package_drupal_script.sh
```diff
@@ -144,11 +144,11 @@ cp ${REPO_DIR}/example.gitignore ${STAGING_DIR}/sprint/drupal8/.gitignore
 
 echo "Running composer install --quiet"
 composer install --quiet
+popd >/dev/null
 
 # Copy licenses and COPYING notice.
-cp -r ${REPO_DIR}/licenses "$STAGING_DIR/"
-cp ${REPO_DIR}/COPYING "$STAGING_DIR/"
-popd >/dev/null
```
mradcliffe, line 151: Whoops. Maybe a good test case to assert the package contents?

tests/sane_testbot.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+echo "sanetestbot.sh: Check to see if test machine has what it needs"
```
mradcliffe: Small typo here. Not a big deal.

tests/sane_testbot.sh
```diff
@@ -0,0 +1,19 @@
...
+# choco install -y jq 7zip composer zip (gd and curl must be enabled in /c/tools/php72/php.ini)
+# apt-get install jq p7zip-full
+# git clone git://github.com/bats-core/bats-core; cd bats-core; git checkout v1.1.0; sudo ./install.sh /usr/local
+
+for item in curl jq 7z composer perl bats; do
+    command -v $item >/dev/null || ( echo "$item is not installed" && exit 2 )
+done
+docker run --rm -t -v "/$PWD:/junk" busybox ls //junk >/dev/null || ( echo "docker is not running" && exit 3 )
+
+echo "Testbot appears to be sane"
```
mradcliffe: Testbots can be sane? ;-)

mradcliffe (commented):
:+1: on moving into a tests directory.

It might be worth it to add a small test to assert the licenses directory since that was a bug fix, right?

Edit: sorry, saw some emails come through and linked directly to the changeset without noticing the label.

tests/sane_testbot.sh
rfay: Fixed.

package_drupal_script.sh
rfay: I added assertion on these items. Will keep thinking about the rest of the contents, but they'd all cause other failures so it doesn't seem as bad.

rfay:
I think this is ready, would love it if you could take a look @mradcliffe. Wow, did that take up the day. It wasn't supposed to be like that.

sprint/start_clean.sh
```diff
@@ -44,11 +44,8 @@ done
 ddev config --docroot drupal8 --projectname sprint-[ts] --projecttype drupal8
 
 ddev start
-ddev exec git fetch
-ddev exec git reset --hard origin/8.7.x
-ddev exec composer install
-ddev exec drush si standard --account-pass=admin --db-url=mysql://db:db@db:3306/db --site-name="Drupal Sprinting"
-ddev exec drush cr
+ddev exec bash -c 'git fetch && git reset --hard origin/8.7.x && composer install && drush si standard --account-pass=admin --db-url=mysql://db:db@db/db --site-name="Drupal Sprinting" && drush cr'
```
mradcliffe, line 48: #77 would probably be important since this is long-running, but not a blocker for this PR.

mradcliffe (commented):
I'm downloading the build artifacts to confirm, and then I think this PR is good.

sprint/start_clean.sh
rfay: Shoot, thought that was in. This particular line actually generates enough output that people will understand that things are going on. But I think an explanatory line would be good here.

mradcliffe (approved):

mradcliffe:
I probably won't have time to focus on any PRs until late in the day today.

PR #82, opened 2018-08-19
Override version of ddev to be used during tests

rfay:
We need a couple of things for valid tests, and don't have them before this PR:
* During tests, we need to install ddev on all platforms even if it already exists
* the ddev we install needs to be the one used.
* If we install for tests it shouldn't be in the normal path.
* It shouldn't require sudo

So that's the intent here. It's done with an environment variable to override the destination for installation.

This can't be reviewed really until rebased and tested after https://github.com/drud/quicksprint/pull/81 goes in.


package_drupal_script.sh
```diff
@@ -118,7 +118,7 @@ ${SHACMD} -c "$SHAFILE"
 popd >/dev/null
 
 # Download the ddev tarball/zipball
-for item in macos linux windows_installer; do
+for item in macos linux windows windows_installer; do
```
mradcliffe: Probably a merge conflict coming up here.

mradcliffe (approved):
I checked out the branch locally and confirmed that install_ddev is no longer in the repository manually.

There might be a merge conflict with #81, but approved until that point.

PR #84, opened 2018-08-21
Minor docs improvements and script improvement

rfay:
In doing a manual Windows test yesterday I found some minor problems (and a script #fail). This PR is the result of those. I didn't get all the way through though, but I think this should be harmless.

### DIFF
--- a/SPRINTUSER_README.md
+++ b/SPRINTUSER_README.md
@@ -33,7 +33,7 @@ This directory contains tools to get you started contributing to Drupal 8:
 <a name="extract-files"></a>
 ### 2. Extract drupal_sprint_package directory
 
-Extract the drupal_sprint_package-RELEASE.zip file, and open or browse to the "drupal_sprint_package" directory.
+Extract the drupal_sprint_package.RELEASE.zip file, and open or browse to the "drupal_sprint_package" directory.
 
 <a name="install"></a>
 ### 3. Install Docker and Other Requirements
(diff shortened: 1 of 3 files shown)


SPRINTUSER_README.md
```diff
@@ -50,19 +50,25 @@ Find the Docker installer for your Operating System underneath the **installs**
  MacOS | Docker CE | Docker.dmg
  Linux | Docker CE, docker-compose | See [Linux instructions](https://docs.docker.com/install/#docker-ce)
 
-**⚠️ Docker Toolbox users**: Please install "Git for Windows" and "VirtualBox" as needed during the install process.
+**⚠️Docker Toolbox users**: Please click the checkboxes to install "Git for Windows" and "VirtualBox" as needed during the install process.
 
-**⚠️ Docker CE only** Start Docker
+**⚠️Docker Toolbox Users must start "Docker Quickstart Terminal" to start docker running.
```
mradcliffe: I initially didn't have this because it happens for these users in step 4 opening a terminal window, but it doesn't hurt.

mradcliffe: Missing a corresponding `**` after "Users" to terminate the bold otherwise the line begins with a literal `**`.

SPRINTUSER_README.md
```diff
@@ -50,19 +50,25 @@ Find the Docker installer for your Operating System underneath the **installs**
...
 
-**⚠️ Docker CE only** Open Docker preferences and set docker memory allocation to 3.0GB or greater in the Advanced section. This is required to start containers.
```
mradcliffe: Isn't this still a good (or necessary) step?

mradcliffe (changes requested):
One question, one formatting fix needed in docs.

SPRINTUSER_README.md
rfay: I use the default 2GB for everything I do, and the testbots also do. The increase was for the hungry IRC and code editor containers. I often run 4-5 projects at a time.

rfay: Note that it is really necessary for Windows users to share the drive(s) though.

mradcliffe: Yes, that's why I'm relegated to Docker Toolbox on this windows desktop I'm testing with :(

rfay: Tell me more - You can't share drives, but things work ok with docker toolbox anyway? Is that a path for people with locked-down computers or something?

install.sh
```diff
@@ -87,24 +87,24 @@ if [ ! -z "$TARBALL" ] ; then
     else
         # Calling script may have already set DDEV_INSTALL_DIR, otherwise we respect and use it.
         # Otherwise, try to use /usr/local/bin and then /usr/bin
-        if [ ! -z "${DDEV_INSTALL_DIR}" ]; then
+        if [ ! -z "${DDEV_INSTALL_DIR:-}" ]; then
```
mradcliffe, line 90: I ran into this on the build on master (ddev.exe exists in my path in Git-Bash) as I installed it in step 3 and this is step 5. 

```
\033[33mPLease use the ddev_windows_installer provided with this package to install ddev\033[0m
./install.sh: line 85: DDEV_INSTALL_DIR: unbound variable
```

mradcliffe (commented):
Found issue with installing ddev step before running install.sh

SPRINTUSER_README.md
mradcliffe: Yes, Docker Toolbox works fine even though I don't have access to File & Print Sharing, but I do have enough Admin. access to run installers.

install.sh
mradcliffe: Ah, it's because TARBALL exists for the test bot, but we don't want it in the real case.

SPRINTUSER_README.md
rfay: I suspect that you might be one of many then. We may want to explain that option to people in your position. Fascinating to know!

install.sh
```diff
@@ -126,3 +126,7 @@ ${GREEN}
 ######
 ${RESET}
 "
+
+if ! command -v ddev && [ ${OS} = "MINGW64_NT-10.0" ] ; then
```
mradcliffe: Should the output of `command -v ddev` be wrapped in `> /dev/null` like the other commands sothat it doesn't print out "/c/Program Files/ddev/ddev" or "/usr/local/bin/ddev"?

install.sh
mradcliffe: Patched install.sh with the commit changes and it worked for me.

mradcliffe (commented):
Fix worked for me.

Pending question.

install.sh
rfay: Yep. So many things.

SPRINTUSER_README.md
rfay: Fixed.

rfay:
I think everything is ready here @mradcliffe - and so we can get this one in and then the 7zip removal.

mradcliffe (approved):

PR #85, opened 2018-08-22
Switch to parallel build/artifacts in circleci

rfay:
Just learned how to parallelize the test and artifacts creation in circleci, so doing that here. Sometimes you just don't want to wait for the tests to complete to work with the artifacts.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,107 +1,170 @@
-version: 2
+version: 2.0
+
 jobs:
-  "package_build":
+
+  "buildjob":
     machine:
-      image: image: circleci/classic:201808-01
-    working_directory: ~/quicksprint
-    environment:
-      TERM: vt100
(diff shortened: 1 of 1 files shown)


mradcliffe (changes requested):
https://github.com/drud/quicksprint/blob/master/tests/test_drupal_quicksprint.sh#L36

I think this needs to be changed to use `/tmp/drupal_sprint_package` rather than the tarball since it might not exist yet, and not cleanup that directory in the test run.

rfay:
I didn't quite understand what @mradcliffe - When this starts working the package/build happens before the test and before the artifact creation. So that should work shouldn't it?

rfay:
You can see and look at the *workflow* builds for this in https://circleci.com/gh/drud/workflows/quicksprint/tree/pull%2F85

circleci_-_circleci

mradcliffe (approved):

PR #86, opened 2018-08-22
Remove 7zip requirement (and bundling)

rfay:
Since git-bash now bundles xz, we don't have to ship 7zip any more and the code is simpler. 

This removes 7z/7zip/p7zip.


install.sh
```diff
@@ -44,15 +44,11 @@ ${RESET}"
 
 echo ""
 echo "Installing docker images for ddev to use..."
-if [ ${OS} = "MINGW64_NT-10.0" ] ; then PATH="./bin/windows:$PATH"; fi
 
-
-if command -v 7z >/dev/null; then
-    7z x ddev_tarballs/ddev_docker_images.${DDEV_VERSION}.tar.xz -so | docker load
-elif command -v xzcat >/dev/null; then
+if command -v xzcat >/dev/null; then
```
mradcliffe, line 48: I'm not sure this is right. I didn't see xzcat bundled with Git-Bash (gzcat, bzcat, zcat are). I only found that `tar` supported xz.

mradcliffe: I'll have to check tomorrow as I don't have access to a Windows machine.

mradcliffe (commented):
I think 7zip is still necessary on Windows.

install.sh
rfay: Yes, xzcat is in git 2.18.0, yay!

rfay: Looks like from the release notes they added it (and tar capability) in 2.16.2.

rfay:
I worked through this with a full test on Windows 10 Pro. 
* Downloaded the package from here (tar)
* Uninstalled docker, ddev, and git. 
* Checked for extra installs of 7zip or xz to remove. 
* Did the required installs of git, docker, ddev
* Shared the C drive in docker.
* Ran install.sh
* Ran start_sprint.sh
* Ran start_clean.sh

Looks like it's all working on that flow.

rfay:
Except for one gotcha, a WIndows thing that everybody should know about when doing support: If the number of hosts assigned to 127.0.0.1 (or any IP) goes over 10, then windows won't resolve any more. So ddev now has `ddev hostname --remove-inactive` to clean up all that extra stuff. In this case, this was the 11th project that I'd spun up.

install.sh
mradcliffe: Oh, I found it. Most of the other utilities are in `/usr/bin` but xzcat is in `/mingw64/bin`.

mradcliffe (commented):
The docs mention to check Git for Windows as needed in the Docker Toolbox install. I think that we should mention to uncheck as they'll be installing Git For Windows later.

mradcliffe (changes requested):
In SPRINTUSER_README.md, section 3.3 Git is now probably important to install before Docker Toolbox and then the note about Git for Windows can mention not to install it since the user already installed it.

rfay:
Reworked the install instructions to put git for windows install first. Good idea.

mradcliffe:
Should this get another test run?

Edit: I think this is good otherwise.

rfay:
I gave it a full run, and it did ok on everything but circleci. If we can't get circleci to behave we'll try to deal with that otherwise. Thanks!

I do hope we can get a couple of people besides us to try the package before Drupal Europe.

PR #87, opened 2018-08-23
Debug apt-get cant lock problem by removing -qq

rfay:
We have a persistent problem on circleci with the new parallel workflow technique. It looks like a Circleci bug, but we can't actually have it, so will have to revert the parallel workflow if it doesn't sort out.


rfay:
This didn't teach us anything. Support request open with Circleci.

PR #89, opened 2018-08-23
Clean up sprint packages on buildkite build

rfay:
The testbot is getting disk space problems fast on win10home, need to clean up.


mradcliffe (approved):
I don't really know build kite, but this is probably the OOS issue, right?

rfay:
Thanks, yes. I'm going to see if I can rebuild that testbot with more space also.

PR #92, opened 2018-08-24
Attempts to check for other supported versions of Windows

mradcliffe:
Not sure if this will work exactly as I'm not on Windows.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -9,7 +9,7 @@ RED='\033[31m'
 GREEN='\033[32m'
 YELLOW='\033[33m'
 RESET='\033[0m'
-OS=$(uname | grep -oE '(Darwin|Linux|MINGW64)' || uname)
+OS=$(uname)
 USER=$(whoami)
 SHACMD=""
 FILEBASE=""
@@ -63,7 +63,7 @@ case "$OS" in
     Darwin)
         TARBALL=ddev_tarballs/ddev_macos.${DDEV_VERSION}.tar.gz
         ;;
-    MINGW64)
+    MINGW64_NT*)
         echo ""
         TARBALL=ddev_tarballs/ddev_windows.${DDEV_VERSION}.tar.gz
         echo "${YELLOW}Please use the ddev_windows_installer provided with this package to install ddev${RESET}"
@@ -123,6 +123,6 @@ ${GREEN}
 ${RESET}
 "
 
-if ! command -v ddev >/dev/null && [ "${OS}" = "MINGW64" ] ; then
+if ! command -v ddev >/dev/null && [ "${OS}" = "MINGW64_NT*" ] ; then
     printf "${RED}ddev has not yet been installed. Please use the ddev_windows_installer to install it${RESET}"
 fi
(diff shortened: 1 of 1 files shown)


rfay:
We do have to have a way to test this, at least trivially testing manually.

mradcliffe:
Looks like it needs work.

rfay:
For testing, you may want these images: https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/ - Parallels makes this super easy (for those on mac). I just grabbed win7, will try to get win8 too, but probably won't have time to do any testing.

rfay:
Windows 7 (32-bit) reports with uname (and thus $OS) `MINGW32_NT-6.1`
Windows 7 (64-bit) reports `MINGW64_NT-6.1`
Windows 8.1 reports with uname `MINGW64_NT-6.3`
Windows 10 reports `MINGW64_NT-10.0`

So basically, we can mostly support `MINGW64_NT`

The Windows 7 instance from microsoft tools though is apparently i386; docker toolbox requires x64. So nobody who comes to a sprint with a 32-bit machine is going to work out I don't think. 

And I can't come up with a way to run Win7 64-bit with nested virtualization, so skunked there.

mradcliffe:
Yep, that's what I was trying to target, but I probably messed up the grep regex. 

Let's see about this one, which tries to simplify the uname string to just MINGW64 so any random Windows version update doesn't bite us.

mradcliffe:
So, maybe the wildcard approach isn't detecting correctly on buildkite, but using grep is fine?

mradcliffe:
> ci/circleci: normal_build_and_test Expected — Waiting for status to be reported 

I think circleci/github is stuck. The job completed successfully.

rfay:
I didn't understand why you gave up on https://github.com/drud/quicksprint/pull/92/commits/8f43e1cc503444ecb55862681b7aaa9a4046f1a6 ? Because the [docker toolbox failure](https://buildkite.com/drud/quicksprint-windows-dockertoolbox/builds/86#6cf3456c-ee30-42d1-942c-44bbc70ec9e0)? that happened long before the install.sh AFAICT (It was cancelled by your next commit... which seemed to have an error in it?)

Restarting those builds for 8f43e1cc503444ecb55862681b7aaa9a4046f1a6

mradcliffe:
I waited for all the builds to finish (and the build kite ones failed for both ) before reverting 
8f43e1c . And I swear it had a red x yesterday along with all the builds of d585a79. :|

rfay (approved):
Well, I like it now :) Thanks, sorry for the back-and-forth. I wouldn't have stretched to try for Win7/8 but glad it kind of works there, that's pretty amazing. I wonder if we should officially support some of that (but it would probably mean we'd have to start testing on old windows, another pain)

IMO you can pull this when you're happy with it.

mradcliffe:
I think it looks good. circleci/github seems to be stuck on expecting a status update when everything's finished again.

rfay:
The overall failure to report of the main job (normal_build_and_test) seems to be just some kind of Circleci bug, but yeah, looks to like everything is fine.

mradcliffe:
I'm ready for this, but I can't merge because of the circleci status check.

PR #95, opened 2018-09-13
Various changes to SPRINTUSER_README

Sutharsan:
Various chnages after testing to iron out some of the hick-ups I encountered. Installed on OS X from memory stick, preparing for DrupalEurope mentored sprint.

### DIFF
--- a/SPRINTUSER_README.md
+++ b/SPRINTUSER_README.md
@@ -13,7 +13,7 @@ This directory contains tools to get you started contributing to Drupal 8:
 * Windows 7 or higher, MacOS El Capitan or higher or a recent/stable Linux distribution.
 * A robust code editor such as Visual Studio Code, Atom, PhpStorm or Netbeans (this may be provided as part of this package).
 
-**⚠️ You can still install and contribute to Drupal even without these requirements using [Drupal's quick-start command](https://www.drupal.org/docs/8/install/quick-start-launch-a-local-demo-version-of-drupal-8-using-4-brief-steps)**!
+**⚠️ If your computer does not match a requirement, try the [Drupal quick-start](https://www.drupal.org/docs/8/install/quick-start-launch-a-local-demo-version-of-drupal-8-using-4-brief-steps)**! Now skip the steps below.
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


SPRINTUSER_README.md
```diff
@@ -57,10 +59,10 @@ Extract the drupal_sprint_package.RELEASE.zip file, and open or browse to the "d
 
 **⚠️Docker for Windows**: You *must* share the C: drive (preferably all drives, or any other drive your home directory may be on) under Docker->Settings->Shared Drives.
 
-**⚠️All users:** additional information is available at [ddev docker instructions](https://ddev.readthedocs.io/en/latest/users/docker_installation/).
```
rfay, line 60: I don't understand the removal of this paragraph - Especially Linux users need more info.

SPRINTUSER_README.md
```diff
@@ -57,10 +59,10 @@ Extract the drupal_sprint_package.RELEASE.zip file, and open or browse to the "d
...
-
 **⚠️Linux users:** You'll probably need the [ddev docker instructions](https://ddev.readthedocs.io/en/latest/users/docker_installation/) to get docker properly set up.
 
+**⚠️All users:** Now start the Docker app you installed. If Docker asks you to login, you can ignore this.
```
rfay, line 64: This actually is misleading. On WIndows and MacOS docker always asks you to log into the *machine* in order to have privs on the filesystem, etc.

rfay (changes requested):
First, thanks for this! I was waiting for the next touch of quicksprint, but wanted to get your thoughts in sooner. 

I did have a couple of requests.

SPRINTUSER_README.md
mradcliffe: I think the intent here is that Docker for * now is persistent in asking the user to create a Docker Hub user account (which isn't necessary). Some users get stuck on this because they think they have to create one and login to use Docker.

rfay: I do think that's the intent (although I haven't noticed it being obnoxious). But it's important to make the differentiation between logging into the machine (for docker install) and logging into dockerhub.

rfay:
Thanks so much for this contribution. In the interest of moving forward your changes were incorporated in the now-merged #96 . Looking forward to more contributions in the future @Sutharsan

PR #96, opened 2019-01-12
Clarifies sprint user docs that confused users

mradcliffe:
Making review changes from #95.

### DIFF
--- a/SPRINTUSER_README.md
+++ b/SPRINTUSER_README.md
@@ -59,9 +59,11 @@ Extract the drupal_sprint_package-RELEASE.zip file, and open or browse to the "d
 
 **⚠️Docker for Windows**: You *must* share the C: drive (preferably all drives, or any other drive your home directory may be on) under Docker->Settings->Shared Drives.
 
-**⚠️Linux users:** You'll probably need the [ddev docker instructions](https://ddev.readthedocs.io/en/latest/users/docker_installation/) to get docker properly set up.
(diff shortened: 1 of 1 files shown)


rfay:
Thanks for stepping back in there @mradcliffe ! One of our priorities this year is to have everything buttoned down *before* Drupalcon with some time for burning flash drives.

rfay:
Looks like the testing on Windows will need some love.

mradcliffe:
Thanks. Yes, I'm putting more time back into this with Brian soonish.

mradcliffe:
I can't remember. Does the buildkite configuration run both the package build and the install or only take the artifacts from circleci and run the installer and sprint tools?

rfay:
Builds and installs - you can look at https://github.com/drud/quicksprint/blob/master/.buildkite/test.sh - I'll try to fix it later this week. We have a release tomorrow so will be attending to that first.

mradcliffe:
I tried running package_drupal_script.sh, but I didn't have jq (easily obtained) so it failed immediately. Not sure if that's the issue on buildkite.

mradcliffe:
Packager fails because zip isn't available (git for windows does not package that).

Was buildkite's thing ever working or maybe it was silently failing?

rfay:
Windows successful now (after retries). Not sure what was wrong, but we'll keep an eye on it and if it's a particular testbot, we'll be able to solve it. I reorganized the required checks and retried the circleci build.

rfay (approved):
This looks good to me, thanks. So we should close #95 as everything there is covered?

mradcliffe:
Yes, that's correct. Since I opened it, you can merge.

PR #105, opened 2019-01-18
Minor improvements to sanetestbot to make sure if ddev is installed it's recent

rfay:
This is a testing-only PR; since the install respects and existing ddev (and uses it) we need to have at least reasonably recent ddev on the machine.


mradcliffe (approved):
I tested the script manually on my home machine as I still have v1.0.0 around, and the condition failed as expected and printed out the error message.

PR #102, opened 2019-01-17
Use ddev --project-name; split out actions for better user feedback. Fixes #100, fixes #99, fixes #103

rfay:
I apologize that this PR got out of control. I ended up trying to fix one thing and then another, and of course and just re-familiarizing myself. I think it's improved quite a lot with this PR, but it unfortunately didn't stay with one small change, which was the intent.

* script was using ddev config --projectname, current preferred is ddev config --project-name #100 
* Install components (git fetch/checkout/composer install/drush si) broken out into separate steps to give better user feedback #99 
* Minor cosmetic updates to default config.yaml
* #103: Allow it to keep ddev tarballs around if they're the right version. Circleci won't, they'll be downloaded and built from scratch there. But at least not re-built for tests.

IMO this can be tested by reviewing the build log and installing from the [circle-ci-built package](https://circleci.com/gh/drud/quicksprint/768#artifacts/containers/0) as a test.

PR #107, opened 2019-01-19
Stop using GZIP options variable since it's deprecated, fixes #106

### LINKED ISSUE #106: Fix "gzip: warning: GZIP environment variable is deprecated"
We're seeing this warning in a number of places:
`gzip: warning: GZIP environment variable is deprecated; use an alias or script`

So we should stop using the GZIP environment variable.

rfay:
$GZIP is apparently deprecated, at least with some versions of gzip/tar.

[Stack overflow article](https://superuser.com/questions/305128/how-to-specify-level-of-compression-when-using-tar-zcvf/1396116#1396116) has many solutions, but they vary by platform. I even added one there for macOS, but it doesn't work anywhere else. So I used `    tar -cf - ${STAGING_DIR_NAME} | gzip -9 >drupal_sprint_package.${QUICKSPRINT_RELEASE}.tar.gz` as the most general solution.

### DIFF
--- a/package_drupal_script.sh
+++ b/package_drupal_script.sh
@@ -147,7 +147,7 @@ rm -rf ${STAGING_DIR}/sprint
 
 cd ${STAGING_DIR_BASE}
 if [ "$INSTALL" != "n" ] ; then
-    tar -cf - ${STAGING_DIR_NAME} | gzip -9 >drupal_sprint_package.${QUICKSPRINT_RELEASE}.tar.gz 
+    tar -cf - ${STAGING_DIR_NAME} | gzip -9 >drupal_sprint_package.${QUICKSPRINT_RELEASE}.tar.gz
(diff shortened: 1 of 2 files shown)


mradcliffe (approved):
Yep, this looks good. Piping works for all OSes.

I found some unrelated output from #102 that I didn't notice before (I probably only looked at the test output, not the package build logs)

`ls: cannot access /home/circleci/tmp/drupal_sprint_package/ddev_tarballs/*: No such file or directory`

This isn't affecting anything.

```
if [ -d $ddev_tarballs ] && (ls $ddev_tarballs/* | grep -v ${LATEST_VERSION}) ; then
    rm $(ls $ddev_tarballs/* | grep -v ${LATEST_VERSION})
fi
```

rfay:
Thanks. We might as well fix that. There's enough output here we don't need any extra!

mradcliffe:
I'm creating a PR for it now.

PR #108, opened 2019-01-21
Specify git core.autocrlf=false and core.eol=lf on checkout, fixes #83

### LINKED ISSUE #83: Git for Windows package gitconfig or tell the user what to do?
If we bundle Git for Windows 2.18.0, then we would probably need to tell users the correct install options to make their lives easier when generating patches (or maybe ship a .gitconfig that specifies the correct behavior).

As I write this up, I think that it makes the most sense to configure the Drupal repository during package build, and then we don't need to worry about global preferences the user sets during the Git for Windows installer.

rfay:
Discussion in OP #83.

The clone here will just specify for the *drupal* checkout core.autocrlf=false and core.eol=lf, which are best for any normal situation. 

This patch does not change the user's global configuration.

Test:
After start_clean.sh, cd into the "drupal" directory and `git config core.autocrlf` (should be false) and `git config core.eol` (should be lf). Most important to test this on Windows, but it will be the same everywhere.

### DIFF
--- a/package_drupal_script.sh
+++ b/package_drupal_script.sh
@@ -4,6 +4,9 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
+# Base checkout should be of the 8.7.x branch
+SPRINT_BRANCH=8.7.x
+
 # This makes git-bash actually try to create symlinks.
 # Use developer mode in Windows 10 so this doesn't require admin privs.
 export MSYS=winsymlinks:nativestrict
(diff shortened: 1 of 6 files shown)


package_drupal_script.sh
```diff
@@ -127,7 +127,7 @@ popd >/dev/null
 
 # clone or refresh d8 clone
 mkdir -p sprint
-git clone --quiet https://git.drupal.org/project/drupal.git ${STAGING_DIR}/sprint/drupal8
+git clone --config core.autocrlf=false --config core.eol=lf --quiet https://git.drupal.org/project/drupal.git ${STAGING_DIR}/sprint/drupal8
```
mradcliffe: This seems like a good opportunity for a couple of new test assertions?

```
cd ${SPRINTDIR}/${SPRINT_NAME}/drupal8
echo "# Test git configuration (${git config core.autocrlf})" >&3
AUTOCRLF=$(git config core.autocrlf)
[ "$(AUTOCRLF)" = "false" ]
```

mradcliffe (changes requested):

mradcliffe:
I'm not sure if the change is related to the Windows build failure.

package_drupal_script.sh
rfay: Great idea.

rfay:
@mradcliffe I think it's ready for another look, thanks.

mradcliffe (approved):
Looks good to me.

PR #109, opened 2019-01-22
Uses find to remove old ddev_tarballs

mradcliffe:
Follow-up from #102:

`ls: cannot access /home/circleci/tmp/drupal_sprint_package/ddev_tarballs/*: No such file or directory`

My solution was to rework as a find command. This is available in git-bash, but I haven't tested manually yet on linux or windows (theoretically the same with GNU find). I was lazy and didn't pipe to xargs since it should be a low number of files.


rfay (approved):
This is a fine technique, thanks. Feel free to pull when you're happy with it.

PR #113, opened 2019-01-29
Add silent windows install to install.sh, fixes #112

### LINKED ISSUE #112: Auto-install using the ddev_windows_installer on Windows
We can now do a silent install with the ddev_windows_installer*.exe, so we should do that; it's not intrusive, and makes one less thing to stop for in the install process.

rfay:
We've been counting on people to manually install the ddev windows install, and there's no reason we can't do it for them.
This:
* Does ddev_windows_installer with silent installation on Windows.
* This probably will need tweaks to pass tests, because it requires privilege escalation. 
* Adds an environment variable intended for test-time only, allowing skipping the painful slow image install. Another way to do this would be to query docker for existing images, but it would seem to add complexity we don't need.

This should be manually tested on Windows. Recent artifacts at https://circleci.com/gh/drud/quicksprint/882#artifacts/containers/0

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -64,6 +64,11 @@ case "$OS" in
         ;;
     MINGW64_NT*)
         echo ""
+        # Assume if DDEV_INSTALL_DIR is set that we *do* need ddev on Windows, install it.
+        # Otherwise, we'll do the install using the installer below.
+        if [ ! -z "${DDEV_INSTALL_DIR:-}" ]; then
+            TARBALL=ddev_tarballs/ddev_windows.${DDEV_VERSION}.tar.gz
+        fi
         ;;
     *)
         printf "${RED}No ddev binary is available for ${OS}${RESET}\n"
--- a/install.sh
+++ b/install.sh
@@ -64,8 +64,6 @@ case "$OS" in
         ;;
     MINGW64_NT*)
         echo ""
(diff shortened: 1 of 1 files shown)


mradcliffe:
I'll do a manual test of this tomorrow:

1. With sprint package (and ddev) already installed to make sure nothing breaks there.
2. Without ddev.

mradcliffe:
I was able to test a [build 896](https://896-123502111-gh.circle-artifacts.com/0/artifacts) out today, barely, and didn't run into any issues with ddev install on Windows 10. I still got a UAC prompt for installation so I'm not sure if that's "silent" or not.

rfay:
Thanks @mradcliffe - It does in fact do the UAC prompt, which it should. But the tests skip over that because they have an existing ddev.

I guess it's not fair to call it silent, as it requires the user do click. What do you think? Do we need some output right before it? Maybe a Y/N prompt?

mradcliffe (approved):
I'm sorry for not getting to this earlier this week, and I hope it didn't disrupt your plans too much.

> I guess it's not fair to call it silent, as it requires the user do click. What do you think? Do we need some output right before it? Maybe a Y/N prompt?

I don't think that the UAC prompt affects anything, and works as expected. I didn't know what silent installer meant.

The code changes make sense.

Issue #13, opened 2018-03-13
Support TYPO3 cms on quicksprint


cyberswat:
We have had requests to support TYPO3 in addition to the default drupal 8.  This request would also apply to the other CMSs supported by DDEV, but we've been asked about TYPO3 first.

horncologne:
TYPO3 CMS core team is getting together 2018.Mar.15 for a sprint. They can get people up to speed using the T3C quickstart docs, which were recently updated. Can we make Sprint-in-a-pocket USB installation stick for TYPO3 CMS happen in time?

rfay:
Wonderful! Might be able to do it.

rfay:
I assume you'd want latest TYPO3 checked out to start?

core team probably has different requirements from a normal public sprint, where people are just learning about everything. 

The reality the existing package would probably do the job with dev pros like that, it just does a bit more than you need by preparing a drupal install. I'll take a look.

horncologne:
Thanks, @rfay. I'm hooking you up with a sprint lead in T3 Slack ...

rfay:
Discussion in slack discovered that this is not needed for the core sprint this week, which has just 10 people. DevDays this summer will be a better candidate. So this can hibernate for a while.

rickmanelius:
Perfect. That certainly helps scope/prioritize things a bit. For 10 individuals, it should be pretty easy to simply walk through the installation and quick start guides as long as bandwidth isn't an issue. Thanks for helping, @rfay!

mradcliffe:
I think it makes sense at this point to move packaging script to a scripts directory, and then have one named for each CMS/project.

Then CI/manual packager can package releases for various sprint tools needed for sprint leads of different CMSes? For instance in Travis, we could add the following. 

```
env:
  - PACKAGE_SCRIPT=scripts/package_drupal.sh
  - PACKAGE=scripts/package_typo3.sh
  # - PACKAGE=scripts/package_wp.sh
```

There probably is an equivalent thing in Circle.

rfay:
Works for me @mradcliffe . However, we will now be able to let the Drupal version mature for a bit before doing anything more.

mradcliffe:
[...]

PR #116, opened 2019-02-09
Adds ide container back and refactors to account for #13

mradcliffe:
This allows us to optionally add in the IDE container back.

I got halfway doing this weeks ago when I realized that it made sense to start doing #13 as well.

I changed to a generic "package.sh" and provided a bit of getopt (should be POSIX compliant) instead of Y/N options (except for the IDE thing, which is Drupal specific). Let's see how the tests handle this, but there shouldn't be any different output.

Things that I think are still missing in this PR:

- [ ] Merge in master when all open PRs for the next release are complete.
- [ ] Might need some further documentation changes/testing with regard to Markdown links.


mradcliffe:
I changed the title because it doesn't solve #13, but it prepares for it.

install.sh
```diff
@@ -46,14 +47,33 @@ echo ""
...
-    gzip -dc ddev_tarballs/ddev_docker_images*.tar.xz | docker load
+    XZCAT="gzip -dc"
+    ${XZCAT} ddev_tarballs/ddev_docker_images*.tar.xz | docker load
 else
     printf "${YELLOW}Unable to load ddev_docker_images. They will load at first 'ddev start'.${RESET}\n"
 fi
 
 
+# Allows the user to choose whether they need an editor with PHP-capabilities.
+if [ -f ddev_tarballs/docker_additions.tar.xz ]; then
```
mradcliffe: I need to figure out how to do something similar in package.sh here in install.sh.

rfay:
I'm leaving for a month on a road trip on Thursday, so kind of backed up and don't plan to take this on today or tomorrow, want to get a better release out first. (I'll be available on road trip, working some)

rfay:
Could you rebase this please? I'll try to take a look soon. Looking forward to you having a try on the other one as well. Thanks!

framework/drupal/docker-compose.ide.yml
```diff
@@ -0,0 +1,24 @@
+version: "3"
+
+# Adds Cloud9 IDE.
+services:
+  ide:
+    container_name: ddev-${DDEV_SITENAME}-ide
+    image: briangilbert/cloud9-alpine:20180318
```
rfay, line 7: Is there a way we could use the upstream cloud9? This one is likely to be unmaintained, and the tag is already old obviously.

Also, I'm not sure how big this is, but would love to have running this be optional. Just have a script that runs it with the mount, rather than attaching it to ddev?

rfay (changes requested):
I was quite surprised to see this take on both the IDE container and the framework reorg. Could we please do them in two separate PRs? Thanks.  Also note the request to see if we can use upstream cloud9 IDE.

rfay:
Was sad to see this introduced at Drupalcon without getting merged or tested or discussed, but for it to go forward, it still needs those things. We could begin with a rebase..

rfay:
@mradcliffe I'd sure love to see this get in well before Amsterdam, if you could rebase, etc. 

Do you have anything else that needs to be done for Amsterdam?

Did people use the IDE container at Seattle?

PR #117, opened 2019-02-10
Error: Accidental commit of ugly debugging code

rfay:
I accidentally committed debugging code which prevented install.sh from ever installing the docker images. Ooops. This removes that debugging code. Ironically, I explicitly added an environment variable to control that behavior for testing, so didn't need to have done that in the first place.


mradcliffe (approved):

rfay:
I guess another release :). That was a pretty bad mistake, and not that noticeable if not paying attention. I was installing on a brand new clean  system (chromebook) and it went way too fast.

PR #120, opened 2019-02-13
Consolidate start_sprint.sh and start_clean.sh , fixes #101

### LINKED ISSUE #101: Simplify installation flow: Just have them run one basic script
IMO the current flow is too complex; there are separate scripts:
* install.sh
* start_sprint.sh
* start_clean.sh

for doing the basic install, then creating a work area, then starting up in the work area. I maintain this is more complex than it needs to be, so will experiment with trying to get it down to just two operations.

rfay:
Per OP #101, use just one script, start_script.sh.


DRUPAL_SPRINTUSER_README.md
```diff
@@ -89,10 +89,7 @@ Linux | Docker CE | Your preferred terminal application
    * Example: Run `cd ~/Desktop/drupal_sprint_package`
 2. Run the `install.sh` command and follow the prompts.
    * Example: `./install.sh`
-3. Follow the instructions that print out at the end of the previous command to create a "sprint instance".
+3. Follow the instructions that print out at the end of the previous command to create a sprint instance.
```
mradcliffe, line 92: Change looks good.

start_sprint.sh
```diff
@@ -6,37 +6,64 @@ set -o errexit
...
+# ${GREEN}Mailhog:    ${YELLOW}http://sprint-${TIMESTAMP}.ddev.local:8025/${GREEN}
 #
-# Your Drupal 8 instance is now ready to use, 
-# execute the following commands in terminal 
-# to start a Drupal 8 instance to sprint on:
+# phpMyAdmin: ${YELLOW}http://sprint-${TIMESTAMP}.ddev.local:8036/${GREEN}
 #
-# ${YELLOW}cd ${SPRINTNAME}${GREEN}
-# ${YELLOW}./start_clean.sh${GREEN}
+# Chat:       ${YELLOW}https://drupal.org/chat to join Drupal Slack or drupalchat.eu!${GREEN}
```
mradcliffe: Unrelated to this feature: Drupalchat url changed to `drupalchat.me`.

tests/installation.bats
```diff
@@ -14,9 +14,8 @@ function setup {
     DHOST=127.0.0.1
     # Extract the IP address we need from DOCKER_HOST, which is formatted like tcp://192.168.99.100:2376
     if [ ! -z "${DOCKER_HOST:-}" ]; then DHOST="$(echo ${DOCKER_HOST} | perl -p -e 's/(tcp:\/\/|:[0-9]+$)//g')"; fi
-
-    cd ${SPRINTDIR} && export SPRINT_NAME=$(./start_sprint.sh)
-    cd ${SPRINTDIR}/${SPRINT_NAME} && echo y | ./start_clean.sh
+    cd ${SPRINTDIR} && ./start_sprint.sh
```
mradcliffe, line 17: It's nice that this was the only test change, and I think that means the tests are working out well.

mradcliffe: > docker is running, continuing.
> chmod: cannot access ‘/home/circleci/tmp/drupal_sprint_package/sprint’: No such file or directory

I noticed that the chmod error came back in the test run (or my memory is bad and it hadn't been fixed yet):  https://github.com/drud/quicksprint/blob/9b919d961b4896ded59d4b2ee048d735449d0c7f/tests/test_drupal_quicksprint.sh#L31

I'm not sure why that would come up since there is a `mkdir -p` immediately preceding it.

mradcliffe (changes requested):
I downloaded the no extra installs version and ran through install.sh on my linux box. That worked fine for me. I had to stop at start_sprint.sh because I need to update docker there.

Optional changes:
- [x] drupalchat url change.
- [x] chmod investigation/fix.

mradcliffe (approved):

PR #122, opened 2019-03-27
Refactor into framework directories

mradcliffe:
Splits #116 into framework-agnostic refactor without the IDE addition. This should address review comments in #116 that aren't related to that.

Let's see how the test bot likes it and if I missed anything.


rfay:
Thanks so much for your work on this!

mradcliffe:
Okay, tests passing so this is ready for review and manual testing again.

rfay:
Yay thanks! Did you rebase it? If not, please do that.

mradcliffe:
Sure. I did a merge thinking we could do a squash here since there's so many commits, but I can rebase it down to one commit instead and force push.

rfay:
Needs rebase

mradcliffe:
I apologize for being so late with this. I rebased, but was having trouble on MacOS with curl / libressl timing out today. I pushed the changes to have it run on the test bot while I figure that out.

package.sh
```diff
@@ -4,20 +4,69 @@ set -o errexit
...
+            echo "      --docker         ${GREEN}Includes Docker installers in the package.${RESET}"
+            echo "      --help           ${GREEN}Prints this message and exits.${RESET}"
+            exit
+            shift;
+            ;;
+        --)
+            shift;
+            break
+            ;;
+    esac
```
mradcliffe, line 58: This probably needs a bit of hardening with a default case option so that it doesn't infinite loop. I added something like this to another bash script earlier ago:

```
*)
   printf "${YELLOW}#### WARNING: Ignoring unknown argument ${1}\n${RESET}"
   shift;
   break
   ;;
```

rfay:
I should say that the actual recent interest from TYPO3 community hasn't been there, although I haven't brought it up. Seems they may have a smaller group at "sprints" that doesn't need the extra support quicksprint gives.

mradcliffe:
Well, those ideas didn't work. I really dislike drupal/coder using a ^ for the version. I run into this issue on my own anyway. At first, I didn't want to use discard-changes, but that doesn't seem to work anyway.

I think the permissions changes (in package_drupal_script.sh or maybe the testbot script) setting things to write may be causing this, but I'm not entirely sure. Then again, the permissions could be getting set differently inside the container might be different.

Edit: ran out of time to work on this today.

rfay:
For silly coder I think you just need `ddev composer config discard-changes true`, https://www.drupal.org/project/coder/issues/2939229#comment-13151522

mradcliffe:
Oh, geez, nice catch on forgetting to set the value to true! Whoops. Thank you.

I had a few minutes and tried that locally on my Mac this morning without any issues.

mradcliffe:
Yay, tests are passing at least. I feel similar to when I climbed up to the top of Half Dome when I was a kid and pretty much just collapsed at the top.

 I'm not sure why buildkite is failing after 1 or 2 minutes each time though since I don't have logs for that.

rfay:
Let's close this for now, it's almost a year stale. When things push us to move back to it we can resurrect it or re-do it. Thanks for working on this!

PR #124, opened 2019-03-28
ddev should not hide output, and should prompt for sudo prompt, fixes #123

### LINKED ISSUE #123: ddev start does not provide me a link like the instructions says it would
Platform: MacOS Mojave

I attempted to install this for the first time on my laptop and couldn't get past the ```ddev start``` part that I assume is near the end of the process.  I have installed Docker and used it before on my laptop without issue but I had never used a ddev instance before.

I'm going to let the process go for a long as I can to see if it'll eventually complete.  It's been going for more than 40 minutes so far and it's either not done or stalled.

rfay:
#123 pointed out that on a clean machine that's never run ddev before, a couple of things can happen:

* the sudo prompt may not be obvious (for adding the project hostname to /etc/hosts)
* The prompt to opt in to ddev instrumentation may be hidden

This PR 
* removes the hiding of `ddev start` output that had been added, to make the sudo request more obvious.
* Does a pre-emptive `ddev config global --instrumentation-opt-in=false` to make sure people don't have to even answer the opt-in question that's there if ~/.ddev/global_config.yaml does not exist.
* Adds a little extra prompting about when that sudo might occur.

For manual testing you really only have to test start_sprint.sh, but:
1. Make sure you `rm -rf ~/.ddev`
2. Sudo should be set to always request a password. I added `Defaults        env_reset,timestamp_timeout=0` and changed `%admin         ALL = NOPASSWD: ALL` on my `visudo` to `%admin         ALL = (ALL) ALL`
3. Make sure you remove existing hostnames. You can do this with `ddev rm -a && sudo ddev hostname --remove-inactive`.

### DIFF
--- a/start_sprint.sh
+++ b/start_sprint.sh
@@ -36,9 +36,10 @@ echo "Using ddev version $(ddev version| awk '/^cli/ { print $2}') from $(which 
 
 ddev config --docroot drupal8 --project-type drupal8 --php-version=7.2 --http-port=8080 --https-port=8443 --project-name="sprint-${TIMESTAMP}"
 
+ddev config global --instrumentation-opt-in=false >/dev/null
 printf "${YELLOW}Configuring your fresh Drupal8 instance. This takes a few minutes.${RESET}\n"
-printf "${YELLOW}Running ddev start...${RESET}\n"
-ddev start >ddev_start.txt 2>&1 || (echo "ddev start failed: $(cat ddev_start.txt)" && exit 101)
+printf "${YELLOW}Running ddev start...YOU MAY BE ASKED for your sudo password to add a hostname to /etc/hosts${RESET}\n"
+ddev start || (printf "${RED}ddev start failed.${RESET}" && exit 101)
 printf "${YELLOW}Running git fetch && git reset --hard origin/${SPRINT_BRANCH}.${RESET}...\n"
 ddev exec bash -c "git fetch && git reset --hard 'origin/${SPRINT_BRANCH}'" || (echo "ddev exec bash...git reset failed" && exit 102)
 printf "${YELLOW}Running 'ddev composer install'${RESET}...\n"
--- a/start_sprint.sh
+++ b/start_sprint.sh
@@ -31,10 +31,10 @@ else
     exit 1
 fi
 
-cd "${SPRINTNAME}"
+cd "${SPRINTNAME}/drupal8"
 echo "Using ddev version $(ddev version| awk '/^cli/ { print $2}') from $(which ddev)"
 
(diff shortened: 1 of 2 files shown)


start_sprint.sh
```diff
@@ -31,18 +31,19 @@ else
     exit 1
 fi
 
-cd "${SPRINTNAME}"
+cd "${SPRINTNAME}/drupal8"
```
mradcliffe, line 34: Should we go back to `${SPRINTNAME}` after it's finished so that a user could run start_sprint.sh again via up arrow (say if anything else fails)?

mradcliffe (commented):
Oh, right, I ran into this as well and had people press enter to pop up the sudo prompt.

Good fix so far. See my question regarding where we put the user when finished.

start_sprint.sh
rfay: IMO they want to be in the drupal8 directory. I don't actually understand why the sprint directory isn't a d8 checkout, not sure why there's a drupal8 underneath it, but I didn't want to make a lot of changes.  I always want to be in the project root when working on anything.

mradcliffe: If I'm in drupal8, and something went wrong and I wanted to start_sprint.sh, then I'd need to change back a directory. I think it might be confusing.

rfay: start_sprint.sh would be run from ~/sprint, not from the sprint directory. You're remembering how start_clean.sh worked before.

rfay: It's also worth just teaching people how to use ddev, and I expect that's what will happen. If "something goes wrong" they ought to be able to recover, between git and ddev. Anyway... still listening!

andrewfrench: I think we should be careful about that -- we don't want to cross the line between quickly enabling contributions and getting in the way to the point of needing to learn ddev (_especially_ if there's not a ddev person on-site).

andrewfrench (approved):
I was able to confirm that the output of `ddev start` is no longer hidden and blocking prompts (like sudo permission) are now presented to the user, so ./start_sprint.sh no longer hangs forever.

start_sprint.sh
rfay: Yeah, I wasn't saying that we should require ddev training for everybody, just that starting over from scratch is not what I would be hoping to do as a sprint participant in the land of so-much opportunity to fix things.

rfay:
@mradcliffe I'm going to go ahead and pull this as is, happy to continue the discussion about what directory people should be in when. My preference would be not to have a "drupal8" subdirectory under the sprint-time-date directory.

PR #138, opened 2019-09-21
Testing updates for Quicksprint, tests only

rfay:
I found that I was getting deeper and deeper in a hole trying to sort out PRs in #137, #136, and #135, making changes in each, not good. 

This is aimed at testing only.
* Allow updating the configuration of buildkite testbots with current ddev
* Install ddev in circleci build
* Use homebrew for installs on the circleci build
* Stop installing composer on the host since we'll use `ddev composer` instead.
* Use simpler patterns deleting built artifacts (anticipating the tarball-split release)

### DIFF
--- a/.circleci/circle_vm_setup.sh
+++ b/.circleci/circle_vm_setup.sh
@@ -25,14 +25,18 @@ sudo add-apt-repository \
 sudo apt-get update -qq
 sudo apt-get install -qq docker-ce
 
-# docker-compose
-sudo rm -f /usr/local/bin/docker-compose
-sudo curl -s -L "https://github.com/docker/compose/releases/download/1.23.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
-sudo chmod +x /usr/local/bin/docker-compose
+if [ ! -d /home/linuxbrew/.linuxbrew/bin ] ; then
+    sh -c "$(curl -fsSL https://raw.githubusercontent.com/Linuxbrew/install/master/install.sh)"
+fi
 
+echo "export PATH=/home/linuxbrew/.linuxbrew/bin:$PATH" >>~/.bashrc
 
-curl -sS https://getcomposer.org/installer -o composer-setup.php
-    sudo php composer-setup.php --install-dir=/usr/local/bin --filename=composer
+. ~/.bashrc
+
+brew update && brew tap drud/ddev
+for item in mkcert ddev docker-compose; do
+    brew install $item || /home/linuxbrew/.linuxbrew/bin/brew upgrade $item
+done
 
 # install recent bats bash testing framework
 BATS_TAG=v1.1.0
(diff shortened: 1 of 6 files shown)


.circleci/circle_vm_setup.sh
```diff
@@ -25,14 +25,18 @@ sudo add-apt-repository \
...
 
+echo "export PATH=/home/linuxbrew/.linuxbrew/bin:$PATH" >>~/.bashrc
 
-curl -sS https://getcomposer.org/installer -o composer-setup.php
-    sudo php composer-setup.php --install-dir=/usr/local/bin --filename=composer
+. ~/.bashrc
+
+brew update && brew tap drud/ddev
+for item in mkcert ddev docker-compose; do
+    brew install $item || /home/linuxbrew/.linuxbrew/bin/brew upgrade $item
```
mradcliffe, line 38: Interesting, I've never heard of linuxbrew.

package_drupal_script.sh
```diff
@@ -135,8 +135,13 @@ git clone --config core.autocrlf=false --config core.eol=lf --quiet https://git.
 pushd ${STAGING_DIR}/sprint/drupal8 >/dev/null
 cp ${REPO_DIR}/example.gitignore ${STAGING_DIR}/sprint/drupal8/.gitignore
 
-echo "Running composer install --quiet"
-composer install --quiet
+echo "Running ddev composer install --quiet"
+set -x
+ddev config --project-type=drupal8
+ddev start
+ddev composer install --quiet
```
mradcliffe: Will this still be available locally via the docker volume (can't remember if it's a bind mount) or the bgsync option? If not, it might lead to some unexpected issues if looking at code in a local editor.

rfay: This is the package script, so it's running only on the testbot, and will never be available to a user. The purpose here is to use `ddev composer` instead of host-installed composer to create the initial version of the vendor directory. 

I'm not sure it was a good idea, but because composer 1.9.0 is required for core dev, and since of course it wasn't up to that level on the testbots (but is in the web container) I thought this problem could be eliminated forever by just using `ddev composer`. But the current test failure on Windows may talk me out of that. Waiting to see!

rfay: Oh, forgot to mention, bgsync is long-deprecated (and worked only on macOS). But anyway, this is all about code that gets created, which is bind-mounted from the host.

rfay:
Annoying, failure on Windows: https://www.drupal.org/project/drupal/issues/3057094#comment-13267130 "Skipped installation of bin bin/composer for package composer/composer: file not found in package". This results in a dangling symlink vendor/bin/composer, which results in tar not being willing to untar vendor/bin/composer because it's dangling. 

This looks like
* A core composer problem that ought to be solved.
* A windows tar failure to do what other tars want to do
* Possibly averted by doing the original build with a normal host-side `composer install` instead of `ddev composer install`.

rfay:
I went back to the traditional composer install to work around the drupal bug, https://www.drupal.org/project/drupal/issues/3082866

Thanks for your review of this @mradcliffe - I'm going to go ahead and pull it so it will unblock the other PRs. I'm happy to address any concerns you might have.

PR #143, opened 2019-09-22
Use the wording 'contribution event' instead of 'sprint', fixes #140

### LINKED ISSUE #140: Drupal world is trying to use "Contribution day" or something instead of "Sprint"
Should we change "sprint" in places like "This script will install everything you need to participate in this sprint". It's easy to change there and in the docs.

rfay:
This just changes wording to "Contribution event' where possible, where it doesn't affect logic or require lots of effort to maintain the changes.


AmyJuneH (commented):
contribution user => contributor

PR #145, opened 2019-09-23
Cleanup quicksprint_ddev_tmp after tests

rfay:
The test was failing to clean up a directory it created, possibly creating future chaos. Delete it.


mradcliffe (approved):
Looks sane. The variable is quoted and defined above in the script.

PR #154, opened 2019-10-12
Update default branch to 9.0.x and provide branch-switch script, fixes #153

### LINKED ISSUE #153: Change SPRINT_BRANCH to 9.0.x about Oct 14, 2019
Most opinion seems to be that the default branch checked out for Drupalcon Amsterdam should be SPRINT_BRANCH=9.0.x. 

According to https://www.drupal.org/docs/9/how-drupal-9-is-made-and-what-is-included/when-does-the-dedicated-drupal-9-core-branch-open 9.0.x will be created sometime around October 14, 2019.

rfay:
This does a few minor fixups in preparation for Drupal 9.0.x and Drupalcon Amsterdam:

* Update the default branch for checkout to 9.0.x and include drush 10
* Provide a script in every sprint directory called switch_branch.sh that allows switching, for example to 8.9.x. The Readme.txt in the script directory explains it, but `./switch_branch.sh 8.9.x` or `./switch_branch.sh 8.8.x`
* Change the build to use git.drupalcode.org for the original clone instead of older git.drupal.org
* Change the default PHP version to current stable 7.3

### DIFF
--- a/sprint/Readme.txt
+++ b/sprint/Readme.txt
@@ -19,3 +19,7 @@ ddev help
 For full ddev documentation see https://ddev.readthedocs.io/
 And support on Stack Overflow: https://stackoverflow.com/tags/ddev
 
+If you need to switch Drupal branches, for example to 9.0.x you can
+use the utility switch_branch.sh:
+./switch_branch.sh 9.0.x
+./switch_branch.sh 8.8.x
--- a/sprint/Readme.txt
+++ b/sprint/Readme.txt
@@ -20,6 +20,7 @@ For full ddev documentation see https://ddev.readthedocs.io/
 And support on Stack Overflow: https://stackoverflow.com/tags/ddev
 
 If you need to switch Drupal branches, for example to 9.0.x you can
-use the utility switch_branch.sh:
+use the utility switch_branch.sh. Although the script stashes changes to
+avoid losing your changes, you're best to save them away yourself first.
 ./switch_branch.sh 9.0.x
 ./switch_branch.sh 8.8.x
--- a/sprint/switch_branch.sh
+++ b/sprint/switch_branch.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+# This script allows switching drupal branch, for example, from 8.9.x to 9.0.x
+# or back
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+if [[ $# != 1 ]]; then
+  echo "Please provide a branch to switch to. For example 'switch_branch.sh 9.0.x'"
+  exit 1
+fi
+
+
+target_branch=$1
+
+pushd drupal8
+set -x
+ddev exec "git stash && git reset --hard && git checkout ${target_branch} && git pull && git stash pop"
+ddev composer require drush/drush:^10
+ddev composer install
+ddev exec drush si --yes standard --account-pass=admin --db-url=mysql://db:db@db/db --site-name='Drupal Contribution Time
+'
+set +x
+popd
+
--- a/sprint/switch_branch.sh
+++ b/sprint/switch_branch.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# This script allows switching drupal branch, for example, from 8.9.x to 9.0.x
+# This script allows switching drupal branch, for example, from 9.0.x to 8.9.x
 # or back
 
 set -o errexit
@@ -8,7 +8,7 @@ set -o pipefail
 set -o nounset
 
 if [[ $# != 1 ]]; then
-  echo "Please provide a branch to switch to. For example 'switch_branch.sh 9.0.x'"
+  echo "Please provide a branch to switch to. For example 'switch_branch.sh 8.9.x'"
   exit 1
 fi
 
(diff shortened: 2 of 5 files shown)


rfay:
Note that  patches in https://www.drupal.org/project/drupal/issues/3072702 do not yet produce an installable 9.0.x, so it's up in the air if it will be installable by contribution day.

sprint/Readme.txt
```diff
@@ -19,3 +19,7 @@ ddev help
 For full ddev documentation see https://ddev.readthedocs.io/
 And support on Stack Overflow: https://stackoverflow.com/tags/ddev
 
+If you need to switch Drupal branches, for example to 9.0.x you can
+use the utility switch_branch.sh:
```
mradcliffe: We should add that this will remove any work that they may be in the middle of.

sprint/switch_branch.sh
```diff
@@ -0,0 +1,26 @@
...
+  echo "Please provide a branch to switch to. For example 'switch_branch.sh 9.0.x'"
+  exit 1
+fi
+
+
+target_branch=$1
+
+pushd drupal8
+set -x
+ddev exec "git stash && git reset --hard && git checkout ${target_branch} && git pull && git stash pop"
```
mradcliffe: Should we warn before destroying any changes?

sprint/switch_branch.sh
```diff
@@ -0,0 +1,26 @@
...
+ddev composer require drush/drush:^10
```
mradcliffe: Will drush 10 be incompatible when downgrading from 9.0.x to 8.8.x?

sprint/switch_branch.sh
```diff
@@ -0,0 +1,26 @@
...
+ddev composer install
+ddev exec drush si --yes standard --account-pass=admin --db-url=mysql://db:db@db/db --site-name=\'Drupal Contribution Time\'
```
mradcliffe, line 26: I think this will error out if the site is already installed, which is the default after running start. A couple of ideas:

* Run `sql:drop` first, but only if a site install detected (`core:status --fields= --format=`)?
* Add to the documentation that should run sql-drop first.

mradcliffe (changes requested):
I added some questions about

- drush version compatibility
- removing work
- installing to an existing instance

sprint/switch_branch.sh
rfay: Try it, I'm pretty sure `drush si` has always done a drop, it warns you before you do it if you don't do -y.

sprint/switch_branch.sh
rfay: According to [compatibility chart](http://docs.drush.org/en/master/install/#drupal-compatibility) drush 10 is fine with Drupal 8.4+

sprint/Readme.txt
rfay: This actually tries not to lose work, although I think a warning is reasonable. `ddev exec "git stash && git reset --hard && git checkout ${target_branch} && git pull && git stash pop", saves changes, then restores them after the switch. We all know it might not work out, though, so it's worth mentioning.

sprint/switch_branch.sh
rfay: As above, it's doing it's level best not to destroy changes. And if it fails, the stash is still there. Not sure the level of effort that should be applied to this.

sprint/switch_branch.sh
rfay:
```
$ drush si -y

 // You are about to DROP all tables in your 'db' database. Do you want to
 // continue?: yes.

 [notice] Starting Drupal installation. This takes a while.
```

sprint/switch_branch.sh
rfay: The thing that worries me the most is not losing changes, but introducing the change to composer.json and composer.lock, and of course that getting into a patch.

mradcliffe (approved):
Thank you for answering my questions. It was fairly late when I looked things over, and I forgot drush handled that already.

rfay:
Since 9.0.x is installable now, I went ahead and changed to use it.

mradcliffe (approved):
Changes look good to me.

sprint/switch_branch.sh
```diff
@@ -18,13 +18,11 @@ target_branch=$1
 pushd drupal8
 set -x
 ddev start
-ddev exec "git stash save"
```
mradcliffe: Don't we want at least a stash before checking out the target branch otherwise the checkout could fail?

rfay: Sorry, didn't mean to mark that as resolved.

I tested pretty extensively and the `git checkout` with just a branch (not specifying files) does not overwrite changes. It seemed to work great. I was having trouble with Windows and performance and found that this approach worked wonderfully. I did manually test the switch_branch using this technique. I'm sure it needs a bats test.

tests/installation.bats
```diff
@@ -61,4 +61,9 @@ function teardown {
     CURL="curl --fail -H 'Host: ${NAME}.ddev.site' --silent --output /dev/null --url $URL"
     echo "# curl: $CURL" >&3
     ${CURL}
+
+    echo "# Testing switch_branch.sh"
```
mradcliffe, line 65: Should switch_branch be in its own test case?

mradcliffe: We could do something like `echo "echo 'some debug text';" >> index.php` to test the resetting.

rfay: Yes, switch_branch can be its own, but each test case requires full, full setup, which is why I cheated. Thanks to this review of bats, I know how to do tests with different requirements next time and it will be better :) I' afraid to be working too much longer with this one.

I did manually do a change to README to test and saw the expected results. I agree it's a good test case, but again, can't consider doing everything on this round.

PR #156, opened 2019-10-22
Go back to 8.9.x, scale back switch_branch.sh

rfay:
Unfortunately, 9.0.x and Drush 10.x aren't going to be compatible yet, and we don't know how to do an automated install without drush. 

This PR 
* goes back to using 8.9.x
* not installing site-local drush at all, which prevents problems with the composer.json anyway.
* switch_branch.sh doesn't try to do a drupal install
* switch_branch.sh drops an existing db.

https://drupal.slack.com/archives/C62H9CWQM/p1571676244050500

https://drupal.slack.com/archives/C62H9CWQM/p1571680129057800

### DIFF
--- a/package_drupal_script.sh
+++ b/package_drupal_script.sh
@@ -5,7 +5,7 @@ set -o pipefail
 set -o nounset
 
 # Base checkout should be of the 8.7.x branch
-SPRINT_BRANCH=9.0.x
+SPRINT_BRANCH=8.9.x
 
 # This makes git-bash actually try to create symlinks.
 # Use developer mode in Windows 10 so this doesn't require admin privs.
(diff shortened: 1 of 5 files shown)


mradcliffe:
I think greg is trying to fix core and drush since it's fairly critical. Maybe due date for what to do should be Thursday (latest Friday)?

Options if drush and core aren't ready:

1. Do the code change in this PR: change back to 8.9.x and then communicate best way to get started on 9.0.x since that is going to be a focus for Amsterdam (drop site, switch branch manually, remove vendor, composer install, then re-install from the UI).
2. Not install Drupal automatically, which wasn't something mentors wanted. And would probably be reverted as soon as drush and drupal core composer were working.

rfay:
Interesting that mentors didn't really want automatic install. It absolutely isn't hard for people to do their own web install. I'd be totally open to just removing that part.

mradcliffe:
Sorry, I was too vague.

Option is that we don't install. _But_ mentors had chosen specifically that it would be better to have Drupal installed already.

PR #162, opened 2019-12-11
Remove doubled php installs from ubuntu and homebrew, fixes #161

### LINKED ISSUE #161: Chocolatey failing on Windows Docker CE and Toolbox buildkite
Seemingly simple PR that didn't change anything buildkite or Windows related failed so something is different about chocolatey or the Windows environments:

`ERROR: Cannot bind argument to parameter 'Path' because it is null.` in  `C:\ProgramData\chocolatey\lib\php\tools\chocolateyInstall.ps1'.`

rfay:
Nightly builds have been failing due to apparent differences between the apt install of php and the brew install. Hopefully this minor change to the setup script will help.

### DIFF
--- a/.circleci/circle_vm_setup.sh
+++ b/.circleci/circle_vm_setup.sh
@@ -5,10 +5,8 @@ set -x
 
 # Basic tools
 
-v=php7.2
-sudo add-apt-repository -y ppa:ondrej/php
 sudo apt-get update -qq
-sudo apt-get install -y -qq jq realpath zip ${v} ${v}-bcmath ${v}-curl ${v}-cgi ${v}-cli ${v}-common ${v}-fpm ${v}-gd ${v}-intl ${v}-json ${v}-mysql ${v}-mbstring  ${v}-opcache ${v}-soap ${v}-readline ${v}-xdebug ${v}-xml ${v}-xmlrpc ${v}-zip;
+sudo apt-get install -y -qq jq realpath zip
 
 # Remove any existing docker
 sudo apt-get remove docker docker-engine docker.io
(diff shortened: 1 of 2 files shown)


.circleci/circle_vm_setup.sh
```diff
@@ -34,9 +32,7 @@ echo "export PATH=/home/linuxbrew/.linuxbrew/bin:$PATH" >>~/.bashrc
 . ~/.bashrc
 
 brew update && brew tap drud/ddev
-for item in mkcert ddev composer php docker-compose; do
-    brew install $item || /home/linuxbrew/.linuxbrew/bin/brew upgrade $item
-done
+brew install mkcert ddev composer php docker-compose
```
mradcliffe: Looks like it's failing because it's trying to compile php 7.4 as the default. I guess we should bump from 7.2 to 7.3, but probably not 7.4 at this point?

rfay:
I'm annoyed with how long the homebrew install takes, think I'll use the apt-get install and curl-install composer

rfay:
This will also fix the problem in #161 of broken php upgrade on WIndows (and stays with php 7.3)

rfay:
It ended up really complicated and fragile to get chocolatey to accept pinning to php7.3, so I just fixed all the testbots so they can handle 7.4. The only place this is actually used is in the initial composer install.

rfay:
Yay, we might have it! If you're ok with this @mradcliffe we'll get it in.

mradcliffe:
Yes, this works for me. Did you end up fixing something manually on the buildkite machines to get it to pass on Windows? I don't see any change that would affect that.

rfay:
Yes, there is apparently a bug in chocolatey that prevents major version upgrades to php from succeeding. So I did the manual process to remove the old php and install the new.

PR #168, opened 2020-02-10
Minor maintenance to switch_branch and start_sprint

rfay:
This PR just fixes up some annoying minor problem

* the `drush si` in start_sprint.sh now properly sets the site name. It was being lost before due to quoting issues.
* Minor improvements to switch_branch.sh so it doesn't stumble on the replacement of coder module (which includes its git folder and shouldn't)
* Minor env additions to tests to try to get them to fail more gracefully.

### DIFF
--- a/sprint/switch_branch.sh
+++ b/sprint/switch_branch.sh
@@ -14,12 +14,16 @@ fi
 target_branch=$1
 
 pushd drupal
+STASHNAME=switch-branch-${RANDOM}
 set -x
 ddev start
-ddev exec  "git fetch && git stash save && git checkout origin/${target_branch}"
-ddev composer install
-ddev exec "( git stash apply || true )"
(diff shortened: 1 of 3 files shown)


mradcliffe (approved):
This looks good to me. I think RANDOM is always populated. I tested on Git-Bash for that.

I'm not sure if you wanted to add a test for the script, but there isn't a current test for it.

rfay:
Thanks for taking a look!

Yes, $RANDOM is a standard shell feature from clear back in bourne shell. 

The switch_branch script is tested here, if that's what you mean: https://github.com/drud/quicksprint/blob/6a6727dcea1b3d7613f213ee445c929ef8370b69/tests/installation.bats#L65-L70

mradcliffe:
Ah, it is tested. Cool.

PR #178, opened 2020-07-13
Update DRUPAL_SPRINTUSER_README.md

chrisdarke42:
Updated documentation to match the current contents of the quicksprint releases directory, where 3rd party tools have been moved into their own download, but also have the name 'drupal_sprint_package' when extracted.


chrisdarke42:
I have finished making the documentation changes that will need to be in place for the current version of the tools.

mradcliffe (approved):

rfay:
I added a couple of minor updates. It's called "Docker Desktop" now, a couple of minor things like that. 

I'll open an issue to move this to the README.md, it should be the first thing people see.

PR #194, opened 2021-03-23
Change docker desktop download locations, fixes #193

### LINKED ISSUE #193: quicksprint is getting older versions of docker
Currently the [installs](https://github.com/drud/quicksprint/releases/download/v1.15.2/quicksprint_thirdparty_installs.v1.15.2.tar.gz) in the latest release https://github.com/drud/quicksprint/releases/tag/v1.15.2 is not picking up current docker for mac. I think they moved the link and probably orphaned the one we need. 

This doesn't do much harm for Midcamp, because the docker version we're bundling (3.1.0.5) is better than the ones they've released since. However it could be confusing for people since it immediately wants to update.

rfay:
Docker has changed the download locations for docker desktop and orphaned the original ones, #193, so we end up with older versions bundled with quicksprint.

This changes to the current download links.

I also had to do some buildkite maintenance for this. Since the last round here architecture has become very important, so the testbots are selected by both OS and architecture.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -59,7 +59,7 @@ if [ -z "${QUICKSPRINT_SKIP_IMAGE_INSTALL:-}" ]; then
 fi
 
 TARBALL=""
-case "$OS/$ARCH" in
+case "${OS}/${ARCH}" in
     Linux/x86_64)
         TARBALL=ddev_tarballs/ddev_linux-amd64.${DDEV_VERSION}.tar.gz
         ;;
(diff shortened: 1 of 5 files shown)


install.sh
```diff
@@ -59,7 +59,7 @@ if [ -z "${QUICKSPRINT_SKIP_IMAGE_INSTALL:-}" ]; then
 fi
 
 TARBALL=""
-case "$OS/$ARCH" in
+case "${OS}/${ARCH}" in
```
mradcliffe, line 62: I think this is okay as it should still match the MINGW64_NT* case due to the wildcard ending there.

Thanks for working on this and getting the updated docker.

rfay: This particular change doesn't change any functionality, it's just a more restrictive use of the variables.

PR #196, opened 2021-03-26
Adds instructions for adding Nightwatch capability to local environment

### LINKED ISSUE #192: xdebug instructions
I am having trouble getting xdebug to work with ddev quicksprint.

```
➜  ~ ddev list
NAME                  TYPE     LOCATION                              URL                                          STATUS
sprint-20201209-1650  drupal9  ~/sprint/sprint-20201209-1650/drupal  https://sprint-20201209-1650.ddev.site:8443  running
DDEV ROUTER STATUS: healthy
➜  ~ cd ~/sprint/sprint-20201209-1650/drupal

➜  drupal git:(2946-255) ✗ ddev xdebug status
xdebug disabled
➜  drupal git:(2946-255) ✗ ddev xdebug on
Enabled xdebug
```

Following instructions from: https://ddev.readthedocs.io/en/stable/users/step-debugging/
PhpStorm "zero-configuration debugging" did not work, I don't get my breakpoints to hit.

Tested this by setting a breakpoint at the top of index.php, but could not get this breakpoint to hit.
```
$autoloader = require_once 'autoload.php';
```

I'm using chrome browser with the "Xdebug helper" extension, this works for me usually in my other environments.

I ended up trying the "PhpStorm "Run/Debug configuration" Debugging" instructions from https://ddev.readthedocs.io/en/stable/users/step-debugging/

Now, when I hit the debug button I get this error, because I am hitting the site using http not https:
http://sprint-20201209-1650.ddev.site:8443/?XDEBUG_SESSION_START=17541
The plain HTTP request was sent to HTTPS port, so that gives me a clue how we can fix this ...

I think the problem is something to do with the 8443 port used by quicksprint?

mradcliffe:
Fixes #192

This is probably more of a Proof of Concept that might be closed.

I didn't think we would want to add the chromedriver image as that would increase the package size, but maybe we do want this to be the default experience?

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,3 +85,15 @@ Linux            | Docker CE, docker-compose  | Your preferred terminal applicat
 
 This may take a few minutes and will provide you with a set of URLs and further instructions for using your contribution environment.
 
+#### Optional configuration
+
+To add Nightwatch to your local environment,
+
+1. Open the `examples/commands/web/` directory.
+2. Copy the `nightwatch` and `phpunit` files into the .ddev/commands/web directory of your local instance.
+3. Copy the `docker-compose.testing.yaml` file into the .ddev/ directory of your local instance.
+4. Open a terminal in the drupal directory of your local instance.
+5. Run `ddev restart` if your local instance is running or `ddev start` to start your local instance.
+6. Run `ddev exec yarn install` or change directory into drupal/core and run `yarn install`
+7. Run `ddev exec nightwatch --tags=core` to run all core Nightwatch tests!
+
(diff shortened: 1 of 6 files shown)


rfay (changes requested):
I'm in favor of this, but it will have to have tests (quicksprint tests) that prove that it keeps working as time moves on. Otherwise it won't be maintainable.

rfay:
Shouldn't we be adding Nightwatch capability, not just instructions?

mradcliffe:
I wasn't sure if we wanted to make it the default or not. Is there any risks for performance?

rfay:
Do people usually need nightwatch? If so, I'd say add it. I don't think performance would be affected.

PR #203, opened 2021-09-29
Support M1 mac (and arm64 linux) in quicksprint, fixes #199

### LINKED ISSUE #199: Support Mac M1 Apple Silicon
ddev supports Apple M1 machines... Quicksprint will have to be smart enough to do that as well.

rfay:
- [x] Download named tarballs for amd64 and arm64 images (separated out)
- [x] Download named tarballs for Linux arm64 and Mac arm64 ddev binary
- [x] Detect the correct ddev binary and docker image package to use at build time
- [x] Add tests to make sure this works on arm64
- [x] Add testing capability on mac M1 so the M1 version gets run

### DIFF
--- a/.buildkite/darwin_arm64.yaml
+++ b/.buildkite/darwin_arm64.yaml
@@ -1,6 +1,6 @@
 steps:
 - command: ".buildkite/test.sh"
-  label: "package and test amd64"
+  label: "package and test arm64"
   agents:
   - "os=macos"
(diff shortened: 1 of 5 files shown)


rfay:
Thanks for taking a look @mradcliffe - We should be able to get this in and create a release right away if you don't know of anything else critical that needs to be done.

rfay:
Yay, clean tests!

